PR #4, opened 2019-12-23
Added installation docs about Blackfire

Nyholm:


Nyholm:
Yes, it works. Profiles shows up as expected and no error messages.

PR #6, opened 2019-12-27
Added gmp extension

petercv:
I needed the gmp extension to be able to verify RS-256 JWT tokens issued by Cognito using the web-token/jwt-signature-algorithm-rsa library.

It is probably easier to add the gmp extension to the default bref layer, but for now I've created this extra extension layer. I hope this is the correct way of doing things. 

I've only been able to test the layer with Docker so far, which worked great. I created a Dockerfile with the following contents:

```
FROM bref/php-74-fpm-dev

COPY --from=bref/layers/gmp-php-74:latest /opt/bref/include/gmp* /opt/bref/include/
COPY --from=bref/layers/gmp-php-74:latest /opt/bref/lib64/*gmp* /opt/bref/lib64/
COPY --from=bref/layers/gmp-php-74:latest /opt/bref-extra/gmp* /opt/bref-extra/
```

I was hoping there was an easier way to copy all changes from the layer Docker image, but I haven't found a better solution yet.

### DIFF
--- a/layers/gmp/Dockerfile
+++ b/layers/gmp/Dockerfile
@@ -1,0 +1,34 @@
+ARG PHP_VERSION
+FROM bref/build-php-$PHP_VERSION AS ext
+
+# Fix library path
+ENV LD_LIBRARY_PATH=/usr/lib:/usr/lib64:$LD_LIBRARY_PATH
+
+# Install packages
+RUN yum install -y gmp-devel
+
+# Build PHP extension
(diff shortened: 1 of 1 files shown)


Nyholm:
It seams like this is working =)

Screenshot 2019-12-30 at 16 57 04

petercv:
Nice, thank you!

Nyholm:
I have published the layer now. 

You can try install dev-master of the php package and try out your extension if you want to.

PR #11, opened 2020-01-17
Add layer odbc

ondrajodas:


Nyholm:
Thank you. 

Im not familiar with odbc. Could you tell me what it is? 
Also, it does not look like a PHP extension. How do I use this?

ondrajodas:
@Nyholm ODBC is php extension for connect to database eg. `Snowflake database`
this function is described here:
https://www.php.net/manual/en/function.odbc-connect.php

layers/odbc/Dockerfile
```diff
@@ -0,0 +1,16 @@
+ARG PHP_VERSION
+FROM bref/build-php-$PHP_VERSION AS ext
+
+ENV LD_LIBRARY_PATH=/usr/lib:/usr/lib64:$LD_LIBRARY_PATH
+
+# Install packages
+RUN yum install -y php-odbc
```
Nyholm: Aren’t you missing a pecl install?

ondrajodas: i can't install odbc via pecl, because this ended with error

ondrajodas: image

Nyholm:
With my little understanding, it does not seam that till will allow you to use odbc functions with php with this PR. You only included the library and no way for php to use that library.

ondrajodas:
@Nyholm ok i looking on it

ondrajodas:
I'm no longer need, so i close this PR. thanks for your time

PR #18, opened 2020-02-05
Able to make only specified layer

sabmeua:
I propose changes Makefile to be able to make only specified layer.
It will reduce the  time taken build for contributors.

```bash
$ make layer=pgsql
```
I confirmed that  there are no difference in exported zip files whether or not specifying layer.


Nyholm:
Excellent!

PR #21, opened 2020-02-22
Use async aws

Nyholm:
It is just a dev-package so this has no real effect on the library. 

Async AWS is way lighter than the official SDK. The vendor folder got 19Mb lighter with this change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,9 +5,9 @@
     "require-dev": {
         "symfony/process": "^5.0",
         "symfony/finder": "^5.0",
-        "aws/aws-sdk-php": "^3.130",
         "mnapoli/silly": "^1.7",
-        "mnapoli/silly-php-di": "^1.2"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -5,9 +5,9 @@
     "require-dev": {
         "symfony/process": "^5.0",
         "symfony/finder": "^5.0",
-        "aws/aws-sdk-php": "^3.130",
         "mnapoli/silly": "^1.7",
-        "mnapoli/silly-php-di": "^1.2"
+        "mnapoli/silly-php-di": "^1.2",
+        "async-aws/lambda": "^0.1.0"
```
mnapoli: 🤯

Nyholm: I can only assume that emoji is "my mind in blown". 

See here for a similar PR: https://github.com/brefphp/bref/pull/575

PR #24, opened 2020-04-09
Added pcov

footballencarta:
Added pcov to allow for code coverage data to be collected in build environments.

Tested this on AWS Lambda using the bref PHP 7.4 layer, and seems to work as expected.


Nyholm:
Thank you!

PR #31, opened 2020-04-21
Add imagick with webp support and binding to newer versions of ImageMagick

RikudouSage:
I tested it with php 7.2, 7.3 and 7.4.

What I tested:

- image can be outputted in the webp format
- the method `compositeImageGravity()` exists

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
   - LAYER=amqp PHP="72 73 74"
   - LAYER=blackfire PHP="72 73 74"
   - LAYER=gmp PHP="72 73 74"
+  - LAYER=imagick PHP="72 73 74"
   - LAYER=mailparse PHP="72 73 74"
   - LAYER=pcov PHP="72 73 74"
   - LAYER=pgsql PHP="72 73 74"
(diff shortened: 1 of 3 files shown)


mnapoli:
FYI the extension already exists in Bref. More here: https://github.com/brefphp/bref/issues/630

RikudouSage:
Yeah, but it uses ImageMagick from repo which is kinda outdated.

mnapoli:
By the way how heavy is this layer?

If it is (with the new system library included), maybe it would make sense to deprecate (and eventually remove) the one bundled with Bref.

If not, having this extension here would duplicate what's in the main layer, it wouldn't make sense.

RikudouSage:
Around 6,800 KiB.

RikudouSage:
Around 14 MiB if you also add HEIC/HEIF support.

RikudouSage:
Had to change the URL to a GitHub one since imagemagick.org deletes older builds once a new one is published.

mnapoli:
Oh wow, this is huge indeed. Probably best to move the extension in here then.

I have opened https://github.com/brefphp/bref/issues/636 to deprecate the built-in imagick extension.

PR #33, opened 2020-04-26
add igbinary

themazim:
added igbinary as drop in replacement for the standard php serializer.

Tested with bref php 7.3 layer and bref php 7.4 layer on lambda. Works as expected.


Nyholm:
Thank you

PR #36, opened 2020-05-19
odbc snowflake implementation

Nemo64:
I just had the task to connect to a [snowflake](https://www.snowflake.com) database from a php lambda and therefor had to create a layer to add support for it.

The problem I see with generally supporting it here is that... well the odbc extension itself is probably fine but useless without drivers (unless you really want to use mysql though odbc for some reason).
And snowflake is probably a niche and the driver for it is 160mb alone so making this a catch all layer is also not really possible.

So I don't expect this to be merged like it is but maybe someone else has good ideas or needs exactly this.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -53,6 +53,7 @@ extension=/opt/bref-extra/amqp.so
 | Mailparse  | `${bref:extra.mailparse-php-74}` | `extension=/opt/bref-extra/mailparse.so`   |
 | Memcache   | `${bref:extra.memcached-php-74}` | `extension=/opt/bref-extra/memcache.so`    |
 | Memcached  | `${bref:extra.memcached-php-74}` | `extension=/opt/bref-extra/memcached.so`   |
+| ODBC Snowflake | `${bref:extra.odbc-snowflake-php-74}` | `extension=/opt/bref-extra/odbc.so` |
 | Pcov       | `${bref:extra.pcov-php-74}`      | `extension=/opt/bref-extra/pcov.so`        |
 | PostgreSQL | `${bref:extra.pgsql-php-74}`     | `extension=/opt/bref-extra/pgsql.so`       |
 | Xdebug     | `${bref:extra.xdebug-php-74}`    | `zend_extension=/opt/bref-extra/xdebug.so` |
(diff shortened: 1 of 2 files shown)


Readme.md
```diff
@@ -51,6 +51,7 @@ extension=/opt/bref-extra/amqp.so
 | Mailparse | `${bref:extra.mailparse-php-74}` | `extension=/opt/bref-extra/mailparse.so` |
 | Memcache | `${bref:extra.memcached-php-74}` | `extension=/opt/bref-extra/memcache.so` |
 | Memcached | `${bref:extra.memcached-php-74}` | `extension=/opt/bref-extra/memcached.so` |
+| ODBC | NaN | `extension=/opt/bref-extra/odbc.so` |
```
Nyholm, line 54: Let's rename this to Snowflake and I'll be happy ro merge. 
```suggestion
| Snowflake | `${bref:extra.snowflake-php-74}` | `extension=/opt/bref-extra/snowflake.so` |
```

Nyholm:
Since extensions are put in individual layers, I am happy to merge niche stuff. It doesn't matter too much if there are only a handful users =)

Readme.md
Nemo64: The odbc extension itself is the same for any future odbc layer and you can't load the same extension twice. I would prefer keeping the file `odbc.so`
Also, Snowflake is working on a pdo extension that just isn't ready yet so that would be a better candidate for the "snowflake" name.

How about naming it `odbc-snowflake` and continue calling the file `odbc.so`? Then if there would be another odbc layer, you can simply include the one odbc extension and configure both libraries though an own ini file.

Nemo64:
@Nyholm I renamed the layer to `odbc-snowflake` but I didn't change the extensions `odbc.so` file. If you really want to I'll do that as well but I think it is unlikely that someone gets multiple versions of the odbc extension to work at the same time especially since it consists of more than just the extension so file.
If you need multiple drivers, you would just need to bring your own ini file to configure both at the same time.

I also added/updated the documentation for it a bit with more hints to those ini files.

Nyholm:
Thank you. 

I agree, it is unlikely. Is there a benefit to call it `odbc.so` over `odbc-snowflake.so`?

Nemo64:
No, but it is a bit misleading. It is _the_ odbc extension and there is nothing special for snowflake.
You won't be able to use `odbc-snowflake.so` and a hypothetical`odbc-mysql.so` at the same time since they define the same functions so you would have to choose one (which you could also do by just ordering the layers in a way that the preferred is on top).
And there are also all the unix `libodbc.*.so` files that are shared which I can't rename... It may be possible to compile it statically, I haven't tried.

The only snowflake specific thing is the snowflake client which i put in `/opt/snowflake_odbc/` which contains the ini file and the actual snowflake client so.
If you'd want to use another odbc client at the same time you would have to provide your second client library somewhere with it's own ini and include _the_ odbc extension.

Nyholm:
Okey. Sorry for my ignorance. 

I think I get it now. So this is "pure" obdc extension, and the layer has some extra snowflake stuff. 

I am happy with this. Could you rebase so I can merge?

Nemo64:
I had to figure this out too and you probably don't have a way to even test it so ask away 😉

I rebased everything.

Nyholm:
Thank you

https://hub.docker.com/r/bref/extra-odbc-snowflake-php-74

PR #51, opened 2020-08-22
chore(exts): add redis and mongodb

shouze:
as they've moved out from base layers in https://github.com/brefphp/bref/pull/723

PR #53, opened 2020-08-26
Adding Redis

Nyholm:
I was too quick merging #51

### DIFF
--- a/layers/redis/Dockerfile
+++ b/layers/redis/Dockerfile
@@ -1,0 +1,11 @@
+ARG PHP_VERSION
+FROM bref/build-php-$PHP_VERSION AS ext
+
+RUN pecl install redis
+RUN cp `php-config --extension-dir`/redis.so /tmp/redis.so
+
+# Build the final image from the lambci image that is close to the production environment
+FROM lambci/lambda:provided
+
+# Copy things we installed to the final image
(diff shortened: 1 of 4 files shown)


shouze (approved):

Nyholm:
Oh no. Can you see why Redis is failing?

shouze:
> Oh no. Can you see why Redis is failing?

Yes this is probably because base image used for the build contains AND load redis by default I guess. So at the moment I guess we need to find a way to unload/avoid loading it. Let me find out a solution.

layers/redis/Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG PHP_VERSION
+FROM bref/build-php-$PHP_VERSION AS ext
+
+RUN pecl install redis
```
shouze, line 4: That's it, so at the moment till redis ext gets removed from base image have to add force option:
```suggestion
RUN pecl install --force redis
```

shouze:
@Nyholm ok so now... it's mongo's turn! I've made the same patch https://github.com/Nyholm/extra-php-extensions/pull/1

shouze (approved):

Nyholm:
Thank you

shouze:
> Thank you

You're welcome!

PR #74, opened 2020-09-17
Add Scrypt Module

airios:
This branch adds [Scrypt-PHP](https://github.com/DomBlack/php-scrypt).


Nyholm:
I think this looks good. Have you tested it?

airios:
@Nyholm  I deployed with 7.4, 7.3, and it worked fine. With 7.2 a few of the symfony packages complained about minimum php support. I checked 7.2 against a local docker build and loading the extension. It worked fine in that environment. Thoughts?

Nyholm:
I’m not sure. Let’s dig in to that when someone has a problem. 7.2 is very old now :)

Thank you

PR #75, opened 2020-09-21
chore: use scratch as final image

shouze:
So, once this one gets merged you can definitely compose extensions more easily.

### BEFORE

As extra extensions docker images are built from `lambci/lambda:provided`, they are pretty big for nothing...
image

And you possibly have to know more in details if the image requires to copy only resulting extension or also other libraries (lib or lib64?).

```Dockerfile
FROM bref/php-74-fpm-dev

COPY --from=bref/extra-redis-php-74 /opt/bref-extra /opt/bref-extra
COPY --from=bref/extra-imagick-php-74 /opt/bref/lib /opt/bref/lib
COPY --from=bref/extra-imagick-php-74 /opt/bref-extra /opt/bref-extra
```

### AFTER

Lightweight extra extension docker image & no more questions.

```Dockerfile
FROM bref/php-74-fpm-dev

COPY --from=bref/extra-redis-php-74 /opt /opt
COPY --from=bref/extra-imagick-php-74 /opt /opt
```

### DIFF
--- a/docs/create_your_own_extension_layer.md
+++ b/docs/create_your_own_extension_layer.md
@@ -1,6 +1,6 @@
 # How to build new layer
 
-To create your own extension layer, compile the extension and any required libraries 
+To create your own extension layer, compile the extension and any required libraries
 with Docker. This guide uses pgsql extension as an example.
 
 First create new extension directory in `layers/` and move there.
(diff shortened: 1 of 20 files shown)


mnapoli:
@Nyholm I'm curious if you see any downside to this? As I told @shouze, we use the lambci images in Bref so that these images are usable (i.e. they can run). But for extensions, they are not intended to be run by themselves, so that change makes sense to me.

Nyholm:
Hm.. Could you really compile PHP itself and an extension on different operating system and still make them run together?

Im happy to do this if it works =)

shouze:
@Nyholm extensions remains compiled in the same image/operating system. This is just on last step of multi stage build that we copy artifacts into an empty image.

Nyholm:
Yeah, but can you compile the image on one OS and use it on another? Or maybe that is what we currently doing?

mnapoli:
The OS they are compiled in is not changing. Take the `ds` extension for example:

```Dockerfile
ARG PHP_VERSION
FROM bref/build-php-$PHP_VERSION AS ext

RUN pecl install ds
RUN cp `php-config --extension-dir`/ds.so /tmp/ds.so

# Build the final image from the lambci image that is close to the production environment
FROM scratch

# Copy things we installed to the final image
COPY --from=ext /tmp/ds.so /opt/bref-extra/ds.so
```

The `FROM scratch` is only for the second container, which is the one that will just contain the `.so` file. The first container in which we compile the extension is not changed, it's still the Bref one.

Nyholm:
> The first container in which we compile the extension is not changed, it's still the Bref one.

Okey. 

-----

Sorry if Im being extra slow today. 

The extension layer only needs to include the new file, right? If the OS does not matter (if it is `scratch` or `lambci/lambda`).. do we need a OS?

Couldn't we just create a layer with only that file?

shouze:
> > The first container in which we compile the extension is not changed, it's still the Bref one.
> 
> Okey.
> 
> Sorry if Im being extra slow today.
> 
> The extension layer only needs to include the new file, right? If the OS does not matter (if it is `scratch` or `lambci/lambda`).. do we need a OS?
> 
> Couldn't we just create a layer with only that file?

This is exactly the purpose of the reserved `scratch` image name in docker: it's an empty image, no OS.

I should have pointed out [the link](https://hub.docker.com/_/scratch) and some explanations first :wink: 

Gophers for example are very familiar whith this kind of image as go produces autonomous binaries by design (no shared libraries).

image

Nyholm:
> This is exactly the purpose of the reserved scratch image name in docker: it's an empty image, no OS.

Oh, =) Im sorry. I confused it with Debian Stretch =)

Yes, this makes a lot of more sense now. haha.

mnapoli:
@Nyholm all good for you if I merge? That will unlock other PRs for bref 1.0.

Nyholm (approved):
Yepp. Looks good. 

Thank you.

PR #37, opened 2020-05-29
WIP: Add GD extension (with freetype support)

drAlberT:

PR #76, opened 2020-09-22
Add GD extension (with freetype support)

shouze:
Supersedes #37 

Example of tested code snippet:
```php
<?php
$jpg_image = imagecreatefromjpeg('sunset.jpeg');
$white = imagecolorallocate($jpg_image, 255, 255, 255);
$font_path = 'Bely Display.ttf';
$text = "This is a sunset!";
imagettftext($jpg_image, 25, 0, 75, 300, $white, $font_path, $text);
imagejpeg($jpg_image, './sunset_with_legend.jpeg');
imagedestroy($jpg_image);
```

Seems to work as expected :tada: 

### Original Image

sunset

### Transformed image

sunset_with_legend

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
   - LAYER=blackfire PHP="72 73 74"
   - LAYER=cassandra PHP="72 73 74"
   - LAYER=ds PHP="72 73 74"
+  - LAYER=gd PHP="73 74"
   - LAYER=gmp PHP="72 73 74"
   - LAYER=igbinary PHP="72 73 74"
   - LAYER=mailparse PHP="72 73 74"
--- a/Readme.md
+++ b/Readme.md
@@ -57,6 +57,7 @@ memory_limit=128M
 | Blackfire  | `${bref:extra.blackfire-php-74}` | `extension=/opt/bref-extra/blackfire.so`   |
 | Cassandra  | `${bref:extra.cassandra-php-74}` | `extension=/opt/bref-extra/cassandra.so`   |
 | DS         | `${bref:extra.ds-php-74}`        | `extension=/opt/bref-extra/ds.so`          |
+| GD         | `${bref:extra.gd-php-74}`        | `extension=/opt/bref-extra/gd.so`          |
 | GMP        | `${bref:extra.gmp-php-74}`       | `extension=/opt/bref-extra/gmp.so`         |
 | gRPC       | `${bref:extra.grpc-php-74}`      | `extension=/opt/bref-extra/grpc.so`        |
 | Igbinary   | `${bref:extra.igbinary-php-74}`  | `extension=/opt/bref-extra/igbinary.so`    |
@@ -72,7 +73,7 @@ memory_limit=128M
 | Scrypt     | `${bref:extra.scrypt-php-74}`    | `extension=/opt/bref-extra/scrypt.so`      |
 | UUID       | `${bref:extra.uuid-php-74}`      | `zend_extension=/opt/bref-extra/uuid.so`   |
 | Xdebug     | `${bref:extra.xdebug-php-74}`    | `zend_extension=/opt/bref-extra/xdebug.so` |
-| Yaml       | `${bref:extra.yaml-php-74}`      | `extension=/opt/bref-extra/yaml.so`   |
+| Yaml       | `${bref:extra.yaml-php-74}`      | `extension=/opt/bref-extra/yaml.so`        |
 
 Note that the "Memcached" layer provides both extension for [Memcache](https://pecl.php.net/package/memcache) and [Memcached](https://pecl.php.net/package/memcached).
 
--- a/layers/gd/Dockerfile
+++ b/layers/gd/Dockerfile
@@ -1,0 +1,54 @@
+ARG PHP_VERSION
+FROM bref/build-php-$PHP_VERSION as ext
+
+# Install required build libraries
+RUN set -ue \
+    ; LD_LIBRARY_PATH= \
+    ; yum -y install \
+        libwebp-devel \
+        libXpm-devel \
+        libpng-devel \
+        libjpeg-devel \
+        freetype-devel \
+;
+
+WORKDIR ${PHP_BUILD_DIR}/ext/gd
+RUN phpize
+ARG PHP_VERSION
+ENV EXT_CONFIGURE_OPTS=" \
+    --with-freetype \
+    --enable-gd \
+    --with-jpeg \
+    --with-png \
+    --with-webp \
+    --with-xpm \
+    --with-zlib \
+"
+RUN if [ "$PHP_VERSION" = "73" ] ; \
+    then \
+        EXT_CONFIGURE_OPTS=" \
+        --with-freetype-dir=${INSTALL_DIR} \
+        --with-gd \
+        --with-jpeg-dir=${INSTALL_DIR} \
+        --with-png-dir=${INSTALL_DIR} \
+        --with-webp-dir=${INSTALL_DIR} \
+        --with-xpm-dir=${INSTALL_DIR} \
+        --with-zlib-dir=${INSTALL_DIR} \
+        " ; \
+    fi; \
+    ./configure \
+        --disable-static \
+        --enable-gd-jis-conv \
+        --enable-shared \
+        ${EXT_CONFIGURE_OPTS} \
+;
+RUN make -j $(nproc)
+RUN make install
+RUN cp "$(php-config --extension-dir)/gd.so" /tmp/gd.so
+
+FROM scratch
+
+COPY --from=ext /usr/lib64/libXpm.so.4 /opt/bref/lib64/libXpm.so.4
+COPY --from=ext /usr/lib64/libwebp.so.4 /opt/bref/lib64/libwebp.so.4
+
+COPY --from=ext /tmp/gd.so /opt/bref-extra/gd.so


layers/gd/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN phpize
+RUN ./configure \
+        --disable-static \
+        --enable-gd \
+        --enable-gd-jis-conv \
+        --enable-shared \
+        --with-freetype \
+        --with-jpeg \
+        --with-webp \
+        --with-xpm \
```
mnapoli, line 27: Any idea if multiple PHP versions are correctly supported? I'm asking because I remember a change in compilation parameters in 7.4: https://github.com/php/php-src/blob/PHP-7.4/UPGRADING#L761-L770

`--with-gd` became `--enable-gd` and `with-png-dir` was removed, which is good here, but what about PHP 7.3?

mnapoli: For reference in the 7.3 build file: https://github.com/brefphp/bref/blob/ffdf7378818d579f6b8748214735a02a060f620a/runtime/base/php-73.Dockerfile#L81-L83

shouze: Good catch!

Only php74 was built in travis, I've recently pushed to build ontop of other versions and it's green but yes, without the png configure flag.

So I will test my snippet code with a png file accross all php versions and make necessary fixes regarding flags if needed.

Readme.md
```diff
@@ -56,7 +56,8 @@ memory_limit=128M
 | AMQP       | `${bref:extra.amqp-php-74}`      | `extension=/opt/bref-extra/amqp.so`        |
 | Blackfire  | `${bref:extra.blackfire-php-74}` | `extension=/opt/bref-extra/blackfire.so`   |
 | Cassandra  | `${bref:extra.cassandra-php-74}` | `extension=/opt/bref-extra/cassandra.so`   |
-| DS         | `${bref:extra.ds-php-74}`        | `extension=/opt/bref-extra/ds.so`          |
+| DS         | `${bref:extra.ds-php-74}`        | `extension=/opt/bref-extra/ds.             |
+| GD | `${bref:extra.gd-php-74}`                | `extension=/opt/bref-extra/gd.so`          |
```
drAlberT, line 60: mind the spaces @shouze

Readme.md
```diff
@@ -56,7 +56,8 @@ memory_limit=128M
 | AMQP       | `${bref:extra.amqp-php-74}`      | `extension=/opt/bref-extra/amqp.so`        |
 | Blackfire  | `${bref:extra.blackfire-php-74}` | `extension=/opt/bref-extra/blackfire.so`   |
 | Cassandra  | `${bref:extra.cassandra-php-74}` | `extension=/opt/bref-extra/cassandra.so`   |
-| DS         | `${bref:extra.ds-php-74}`        | `extension=/opt/bref-extra/ds.so`          |
+| DS         | `${bref:extra.ds-php-74}`        | `extension=/opt/bref-extra/ds.             |
```
Nyholm, line 59:
```suggestion
| DS         | `${bref:extra.ds-php-74}`        | `extension=/opt/bref-extra/ds.so`             |
```

shouze:
@mnapoli I've found the right options for php73 but still fighting with php72, hope to be ready for a merge at the begining of next week :wink:

mnapoli:
@shouze awesome! But what if we skip PHP 7.2? GD is already in Bref 0.5 layers, and here we want it in a separate layer for Bref 1.0 which will only support 7.3+

drAlberT:
I agree too. No need to support php 7.2 in fact

layers/gd/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+;
+RUN make -j $(nproc)
+RUN make install
+
```
drAlberT, line 31: this empty line is not needed in fact

.travis.yml
```diff
@@ -22,6 +22,7 @@ env:
   - LAYER=blackfire PHP="72 73 74"
   - LAYER=cassandra PHP="72 73 74"
   - LAYER=ds PHP="72 73 74"
+  - LAYER=gd PHP="73 74"
```
shouze, line 25: I've removed php 72 support as discussed.

layers/gd/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    fi \
+;
+RUN make -j $(nproc)
+RUN make install
+RUN cp "$(php-config --extension-dir)/gd.so" /tmp/gd.so
+
+FROM scratch
+
+COPY --from=ext /usr/lib64/libXpm.so.4 /opt/bref/lib64/libXpm.so.4
+COPY --from=ext /usr/lib64/libwebp.so.4 /opt/bref/lib64/libwebp.so.4
```
shouze, line 53: Regarding [Amazon Linux 1 packages](https://github.com/lambci/docker-lambda/blob/master/base/packages.txt), we have only the 2 following ones to copy.

shouze:
@drAlberT @mnapoli ok I've pushed and tested it's working with php 73 & 74 :tada: 

@mnapoli and I believe... that I've found a bug related to local dev docker images :thinking: 
Let me explain and correct me if I'm wrong. During my tests, I've used `ldd` to check against shared libraries. Basicaly, I've created this kind of Dockerfile:

```Dockerfile
FROM bref/php-73

COPY --from=bref/layers/gd-php-73 /opt /opt
```

When I ran it, it returned me:
```shell
docker run --rm -it --entrypoint= foo:bar ldd /opt/bref-extra/gd.so
        linux-vdso.so.1 =>  (0x00007ffd4db1f000)
        libX11.so.6 => /usr/lib64/libX11.so.6 (0x00007f971668f000)
        libXpm.so.4 => not found
        libpng12.so.0 => /usr/lib64/libpng12.so.0 (0x00007f9716469000)
        libz.so.1 => /opt/bref/lib/libz.so.1 (0x00007f971624c000)
        libjpeg.so.62 => /usr/lib64/libjpeg.so.62 (0x00007f9715ff9000)
        libwebp.so.4 => not found
        libfreetype.so.6 => /usr/lib64/libfreetype.so.6 (0x00007f9715d5d000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f971598f000)
        libxcb.so.1 => /usr/lib64/libxcb.so.1 (0x00007f9715771000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f971556d000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f971526b000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f9716c35000)
        libXau.so.6 => /usr/lib64/libXau.so.6 (0x00007f9715068000)
```

Do you see the 2 not found libs? So, of course it works like a charm when I define `LD_LIBRARY_PATH`: 
``shell
docker run --rm -it --entrypoint= --env LD_LIBRARY_PATH=/opt/bref/lib64:/opt/bref/lib foo:bar ldd /opt/bref-extra/gd.so
        linux-vdso.so.1 =>  (0x00007fff11fb3000)
        libX11.so.6 => /usr/lib64/libX11.so.6 (0x00007f835e7f4000)
        libXpm.so.4 => /opt/bref/lib64/libXpm.so.4 (0x00007f835e5e3000)
        libpng12.so.0 => /usr/lib64/libpng12.so.0 (0x00007f835e3bd000)
        libz.so.1 => /opt/bref/lib/libz.so.1 (0x00007f835e1a0000)
        libjpeg.so.62 => /usr/lib64/libjpeg.so.62 (0x00007f835df4d000)
        libwebp.so.4 => /opt/bref/lib64/libwebp.so.4 (0x00007f835dcfe000)
        libfreetype.so.6 => /usr/lib64/libfreetype.so.6 (0x00007f835da62000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f835d694000)
        libxcb.so.1 => /usr/lib64/libxcb.so.1 (0x00007f835d476000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f835d272000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f835cf70000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f835cd54000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f835ed9a000)
        libXau.so.6 => /usr/lib64/libXau.so.6 (0x00007f835cb51000)
```

And I guess that it's related to the fact that at the end of the build process of docker images in bref repo [the latest stage copy file into lambci without redefining `LD_LIBRARY_PATH`](https://github.com/brefphp/bref/blob/669d651ac241cec3394a596166ef74f7770b5965/runtime/layers/fpm/Dockerfile#L13-L14).

**Comment by shouze** [comment]
BTW I guess I have to change paths of copied libraries to either /opt/lib (should work both for local development and deployed lambda on AWS) or /opt/bref/lib (should work for deployed lambda on AWS but not for local development)

https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

**Comment by shouze** [comment]
Also, can I add php 80 or it's too early?
Because I've also tested and... it's working like a charm!

File: Readme.md (hunk 3 above)
Comment (mnapoli) [inline Readme.md:59]: There are still changes on that line

**Comment by mnapoli** [comment]
Awesome!

Regarding `LD_LIBRARY_PATH` I would have expected that path to be set by the lambci Docker images 🤔  Because things work on Lambda so I would expect that path to be defined.

You mention you see a problem with `ldd`, but is the extension working?

FYI `/opt/lib` is not writable on Lambda, so we should not use it even for local development. We should target to mirror Lambda's behavior even in Docker images.

> Also, can I add php 80 or it's too early?

Sounds awesome! Does that involve a different Dockerfile? Also I'm not sure if that involves other changes in the scripts of this repository, so it might be worth putting that in a separate PR?

File: Readme.md (hunk 3 above)
Comment (shouze) [reply Readme.md:59]: ho yes, forgot to rebase before push sorry

**Comment by shouze** [comment]
> Awesome!
> 
> Regarding `LD_LIBRARY_PATH` I would have expected that path to be set by the lambci Docker images  Because things work on Lambda so I would expect that path to be defined.

Yes, but it can't define specific bref paths like /opt/bref/lib or /opt/bref/lib64 BTW. I can open an issue on bref repo at least to continue the discussion on that point, WDYT?
It will become a major subject with Amazon Linux 2 because [lib jpeg, png or freetype are no more available for example](https://github.com/lambci/docker-lambda/blob/master/base/base-2/packages.txt).


> You mention you see a problem with `ldd`, but is the extension working?

Yes, but I'm pretty sure that if I try to generate a webp or xpm images it will crash because shared libs can't be find.

> FYI `/opt/lib` is not writable on Lambda, so we should not use it even for local development. We should target to mirror Lambda's behavior even in Docker images.

Sure!

> > Also, can I add php 80 or it's too early?
> 
> Sounds awesome! Does that involve a different Dockerfile? Also I'm not sure if that involves other changes in the scripts of this repository, so it might be worth putting that in a separate PR?

No other changes required into the Dockerfile. And, yes definitely in another PR, maybe a more global one adding php 80 support to every extensions that are already php 80 ready. WDYT?

File: layers/gd/Dockerfile [hunk 7]
```diff
@@ -0,0 +1,55 @@
...
+        libpng-devel \
+        libjpeg-devel \
+        freetype-devel \
+    ; yum clean all \
+;
+
+WORKDIR ${PHP_BUILD_DIR}/ext/gd
+RUN phpize
+ARG PHP_VERSION
+RUN if [ "$PHP_VERSION" = "73" ] ; \
```
drAlberT, line 19: a bit ugly IMHO ... the condition should target only the different stuff .. never repeat the common part .. WYT?

Something like the following (not tested it could contains errors ...)
```
LIB_STR="
    --with-gd
    --with-jpeg
    --with-png
    --with-xpm
    --with-webp
    --with-zlib
" ; \
if [ ... ]; then
    LIBS_STR="
        --with-gd
        --with-jpeg-dir=${INSTALL_DIR}
        --with-png-dir=${INSTALL_DIR}
        --with-xpm-dir=${INSTALL_DIR}
        --with-webp-dir=${INSTALL_DIR}
        --with-zlib-dir=${INSTALL_DIR}
   " 
fi; \
./ configure \
        --disable-static \
        --enable-gd-jis-conv \
        --enable-shared \
;
```

layers/gd/Dockerfile
drAlberT: I'd copy the exact version we built against .. not relying on sys libs that could break something ...

layers/gd/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+        ./configure \
+        --disable-static \
+        --with-gd \
+        --enable-gd-jis-conv \
+        --enable-shared \
+        --with-zlib-dir=${INSTALL_DIR} \
+        --with-jpeg-dir=${INSTALL_DIR} \
+        --with-png-dir=${INSTALL_DIR} \
+        --with-webp-dir=${INSTALL_DIR} \
+        --with-xpm-dir=${INSTALL_DIR} \
```
drAlberT, line 30: please keep the same ordering in both parts, thus minimising errors when making a comparison (Alphabetical ordering would be the best IMHO)

shouze:
@drAlberT 
> I'd copy the exact version we built against .. not relying on sys libs that could break something ...

I'm not sure to understand, can you please ellaborate?

layers/gd/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# Install required build libraries
+RUN set -ue \
+    ; LD_LIBRARY_PATH= \
+    ; yum -y install \
+        libwebp-devel \
+        libXpm-devel \
+        libpng-devel \
+        libjpeg-devel \
+        freetype-devel \
+    ; yum clean all \
```
shouze, line 13: @drAlberT I'm not sure this is usefull?

layers/gd/Dockerfile
shouze: I'm not sure to understand what you mean. If it's about copying ALL the shared libraries I'm not agree at all because as you can see with `ldd` on gd.so: 
```
        linux-vdso.so.1 =>  (0x00007fff11fb3000)
        libX11.so.6 => /usr/lib64/libX11.so.6 (0x00007f835e7f4000)
        libXpm.so.4 => /opt/bref/lib64/libXpm.so.4 (0x00007f835e5e3000)
        libpng12.so.0 => /usr/lib64/libpng12.so.0 (0x00007f835e3bd000)
        libz.so.1 => /opt/bref/lib/libz.so.1 (0x00007f835e1a0000)
        libjpeg.so.62 => /usr/lib64/libjpeg.so.62 (0x00007f835df4d000)
        libwebp.so.4 => /opt/bref/lib64/libwebp.so.4 (0x00007f835dcfe000)
        libfreetype.so.6 => /usr/lib64/libfreetype.so.6 (0x00007f835da62000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f835d694000)
        libxcb.so.1 => /usr/lib64/libxcb.so.1 (0x00007f835d476000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f835d272000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f835cf70000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f835cd54000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f835ed9a000)
        libXau.so.6 => /usr/lib64/libXau.so.6 (0x00007f835cb51000)
```

We should in this case even copy the libc, libdl and other ones and we would probably collide with every other extensions because every of them have links to those libs.

Amazon don't change their base images very often but yes it breaks ABI when it's the case and it should be managed before it happen.

layers/gd/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+;
+
+WORKDIR ${PHP_BUILD_DIR}/ext/gd
+RUN phpize
+ARG PHP_VERSION
+ENV CONFIGURE_OPTS=" \
```
drAlberT, line 19: `PHP_CONFIGURE_OPTS` could be more safe and in line with the line above :)

layers/gd/Dockerfile
drAlberT: as we copy needed files away from the base image it should be not influent in fact. It is a best practice to reduce layers size on build... 

should be pruneable

shouze: Yes but as we are multi stage we don't care here

drAlberT:
> @drAlberT
> 
> > I'd copy the exact version we built against .. not relying on sys libs that could break something ...
> 
> I'm not sure to understand, can you please ellaborate?

I completely forgotted what I meant ... LOL .. never mind it. Sorry :)

shouze:
@drAlberT so I've just pushed with last updates.

drAlberT:
> @drAlberT so I've just pushed with last updates.

Great job. LGTM 👍🏻

mnapoli (approved):
Awesome!

Nyholm:
Thank you

PR #77, opened 2020-09-30
chore(build): allow to override some makefile variables

shouze:
This is pure DX. This way you can, for example:

```shell
layer=gd php_versions="74 80" make
```

Maybe it worth to be mentioned into README? :thinking:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
-php_versions = 72 73 74
-layer = *
+php_versions ?= 72 73 74
+layer ?= *
 parallel = $(if $(shell which parallel),true,false)
 
 define generate_list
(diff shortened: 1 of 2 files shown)


Nyholm:
Yes please"

Nyholm:
But you can do this already, can't you?

shouze:
> But you can do this already, can't you?

Didn't work for me unless I add [the `?=` operator](https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting).

shouze:
> But you can do this already, can't you?

Ok I understand, we don't launch make the same way. I usualy run it with variables defined like env vars, before the make command:
```shell
layer=gd php_versions=74 make layers
```
And this is what my PR solves in fact.


BTW it's already working the way you're used too without the `?=` operator patch.
```shell 
make layers layer=gd php_versions=74
```

Nyholm:
Im happy with adding this. But we should maybe document this (better).

shouze:
> Im happy with adding this. But we should maybe document this (better).

sorry for the delay, done!

Nyholm (approved):

PR #82, opened 2020-10-15
Copy output with docker cp

### LINKED ISSUE #80: docker error: "tar" executable not found in $PATH
Hello!

I am trying to test a new layer before send the PR, but when I execute the ```make layers``` docker shows me an error:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"tar\": executable file not found in $PATH": unknown.
ERRO[0000] error waiting for container: context canceled 
```
I lost some hours trying to understand what is happening. Is only happening to me or maybe is general? I think that some is wrong in the main docker image```bref/build-php-XX``` because seems that ```tar``` command is not available. But I am not sure.

Anybody can help me to fix it with some clue?

Thank you so much

Nyholm:
This will fix #80. 

We cannot use `tar` anymore since it is not available in `scratch`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ layers: docker-images
 			echo "### Exporting $${dir} PHP$${php_version}"; \
 			echo "###"; \
 			cd ${PWD} ; rm -rf export/tmp/${layer} || true ; cd export/tmp ; \
-			CID=$$(docker create --entrypoint=/bin/bash bref/$${dir}-php-$${php_version}) ; \
+			CID=$$(docker create --entrypoint=scratch bref/$${dir}-php-$${php_version}) ; \
 			docker cp $${CID}:/opt . ; \
 			docker rm $${CID} ; \
 			cd ./opt ; \
@@ -83,5 +83,4 @@ publish-docker-images: docker-images
 			echo ""; \
 		done \
 	done
-
 
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ docker-images:
 layers: docker-images
 	PWD=pwd
 	rm -rf export/layer-${layer}.zip || true
-	mkdir export/tmp
+	mkdir -p export/tmp
 	set -e; \
 	for dir in layers/${layer}; do \
 		for php_version in $(php_versions); do \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -36,7 +36,7 @@ docker-images:
 layers: docker-images
 	PWD=pwd
 	rm -rf export/layer-${layer}.zip || true
-	mkdir export/tmp
+	mkdir export/tmp || true
```
shouze, line 39: it's kinda hack but I prefer, WDYT?
```suggestion
	mkdir -p export/tmp
```

Makefile
```diff
@@ -45,8 +45,11 @@ layers: docker-images
 			echo "### Exporting $${dir} PHP$${php_version}"; \
 			echo "###"; \
 			cd ${PWD} ; rm -rf export/tmp/${layer} || true ; cd export/tmp ; \
-			docker run --entrypoint "tar" bref/$${dir}-php-$${php_version} -ch -C /opt . | tar -x ; \
-			zip --quiet -X --recurse-paths ../`echo "$${dir}-php-$${php_version}" | sed -e "s/layers\//layer-/g"`.zip . ; \
+			CID=$$(docker create --entrypoint=/bin/bash bref/$${dir}-php-$${php_version}) ; \
```
shouze, line 48: I was not very hot to run a container but in fact, chances that cp can fail are nearly close to zero, so :100:

Makefile
Nyholm: Thank you

Makefile
Nyholm: We are actually never running the container. We just create it. 

`cp` is not part of the container. It is a docker command. So it will not fail (on docker 1.0 and above.)

shouze: :warning: ho... but no, in fact... have you tried? Because `scratch` image is trully empty and doesn't contain `bash` at all this souldn't work.

shouze: This is why I was looking for `docker save` option, which imply to untar each docker layer (tar) to reconstitute the image filesystem, thus to be able to zip it.
This is a little shell script that need to:
1. list all layers (docker save export tar paths in the manifest, just `cat manifest.json | jq -r '.[0].Layers'` to get them).
2. loop over and untar.

shouze: > We are actually never running the container. We just create it.

ok so this is the beauty of the hack, well done & very subtle :clap:.
So I guess that you can remove `--entrypoint=/bin/bash` and it still work no? :thinking:

Nyholm: > ho... but no, in fact... have you tried?

Yes, but please confirm: 

```
make layers layer=blackfire php_versions=74
```

shouze, line 48: As we don't run the container entrypoint definition seems useless unless it's required by docker create, no?
```suggestion
			CID=$$(docker create bref/$${dir}-php-$${php_version}) ; \
```

Nyholm: > So I guess that you can remove --entrypoint=/bin/bash and it still work no? 🤔

Both yes and no. A container needs an entrypoint to be created. However, we are never using it.. 

I think I've could have done `--entrypoint=/foobar`

Nyholm: https://github.com/brefphp/extra-php-extensions/pull/82#discussion_r505447666

shouze: I've tried and I confirm, it works perfectly! :ok_hand:

shouze, line 48: Indeed I've checked too, this is pure detail but as it's hacky, I suggest to set an entrypoint that explain we should expect bash here, maybe by reusing the `scratch` term?
```suggestion
			CID=$$(docker create --entrypoint=scratch bref/$${dir}-php-$${php_version}) ; \
```

shouze (approved):
LGTM, just a detail regarding the entrypoint

Makefile
Nyholm: Yes, This make sense. 

Thank you

PR #85, opened 2020-10-19
Add sqlsrv extension

atbreuer11:
Using the PDO extension on a lumen project


Nyholm (approved):
Thank you. 

Have you tested this?

atbreuer11:
@Nyholm The PDO extension has been tested and we are using it in production. We don't use the non-PDO extension, but I did test that I could connect to our DB and run a query.

Nyholm:
Awesome. Thanks

PR #87, opened 2020-10-28
Added cache for docker images

Nyholm:
Let's see if this makes things faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ before_cache:
   # Save docker images
   - mkdir -p docker_images
   - docker save -o docker_images/images.tar $(docker images -a -q)
+  - ls -l docker_images/images.tar
 
 before_install:
   - |
@@ -75,4 +76,3 @@ script:
     for php_version in $PHP; do
       tfold "Build $php_version" docker build --build-arg PHP_VERSION=${php_version} .
     done
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ cache:
 
 before_cache:
   # Save docker images
+  - mkdir -p docker_images
   - docker save -o docker_images/images.tar $(docker images -a -q)
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,14 @@ env:
   - LAYER=xdebug PHP="72 73 74"
   - LAYER=yaml PHP="72 73 74"
 
+cache:
+  directories:
+  - docker_images
+
+before_cache:
+  # Save tagged docker images
+  - docker save -o docker_images/images.tar $(docker images -a -q)
+
 before_install:
   - |
     # tfold is a helper to create folded reports (From Symfony)
@@ -56,6 +64,9 @@ before_install:
         (exit $ok)
     }
     export -f tfold
+
+  # Load cached docker images
+  - docker load -i docker_images/images.tar || true
 
 script:
   - cd layers/$LAYER
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ cache:
   - docker_images
 
 before_cache:
-  # Save tagged docker images
+  # Save docker images
   - docker save -o docker_images/images.tar $(docker images -a -q)
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,3 +75,4 @@ script:
     for php_version in $PHP; do
       tfold "Build $php_version" docker build --build-arg PHP_VERSION=${php_version} .
     done
+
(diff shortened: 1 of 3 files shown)


shouze:
> Let's see if this makes things faster

If this try is not concluant, I have an alternative by using docker buildx and yes there's a room of improvement. Gain depends on cache size to initialy download.

Nyholm:
Oh, I am not sure if this will help things at all. 

Feel free to make a PR to improve things =)

.travis.yml
```diff
@@ -39,6 +39,14 @@ env:
   - LAYER=xdebug PHP="72 73 74"
   - LAYER=yaml PHP="72 73 74"
 
+cache:
+  directories:
+  - docker_images
+
+before_cache:
+  # Save docker images
+  - docker save -o docker_images/images.tar $(docker images -a -q)
```
sanmai, line 48:
```suggestion
  - mkdir -p docker_images
  - docker save -o docker_images/images.tar $(docker images -a -q)
```

Nyholm: Thank you

.travis.yml
```diff
@@ -39,6 +39,15 @@ env:
...
   - LAYER=yaml PHP="72 73 74"
 
+cache:
+  directories:
+  - docker_images
+
+before_cache:
+  # Save docker images
+  - mkdir -p docker_images
+  - docker save -o docker_images/images.tar $(docker images -a -q)
```
sanmai, line 49:
```suggestion
  - docker save -o docker_images/images.tar $(docker images -a -q)
  - ls -l docker_images/images.tar
```
Is there anything at all?

sanmai:
Admin can check out if there's anything in the cache here: https://travis-ci.com/github/brefphp/extra-php-extensions/caches

Nyholm:
Hm. 
Screenshot 2020-10-29 at 09 16 01

sanmai:
Another, slightly more complicated option would be to use ccache from a shared volume: https://stackoverflow.com/questions/39650056/using-ccache-when-building-inside-of-docker

shouze:
> Another, slightly more complicated option would be to use ccache from a shared volume: [stackoverflow.com/questions/39650056/using-ccache-when-building-inside-of-docker](https://stackoverflow.com/questions/39650056/using-ccache-when-building-inside-of-docker)

@sanmai I guess using buildx is a better option.

### Buildx setup (can be added into Makefile)

```Makefile
$(HOME)/.docker/cli-plugins/docker-buildx:
	@mkdir -p $(HOME)/.docker/cli-plugins \
	&& curl -qL https://github.com/docker/buildx/releases/download/v0.4.2/buildx-v0.4.2.linux-amd64 > $(HOME)/.docker/cli-plugins/docker-buildx \
	&& chmod +x $(HOME)/.docker/cli-plugins/docker-buildx \
	&& sleep 1

buildx: $(HOME)/.docker/cli-plugins/docker-buildx
	@docker buildx version \
	&& docker buildx create --driver=docker-container --name builds --use > /dev/null || echo "buildx builds node already exists"
```

### Build with buildx (can replace current Makefile build process)

```Makefile
build: PUSH?=
build:
ifndef PUSH
	@docker buildx build \
		-t $(DOCKER_PACKAGE_NAME) . \
		-f ./Dockerfile \
		--platform=linux/amd64 \
		--pull \
		--cache-from=type=registry,ref=$(DOCKER_CACHE_TAG) \
		--cache-to=type=inline,mode=max \
		--load
else
	@docker buildx build \
		-t $(DOCKER_PACKAGE_NAME) . \
		-f ./Dockerfile \
		--platform=linux/amd64 \
		--pull \
		--cache-from=type=registry,ref=$(DOCKER_CACHE_TAG) \
		--cache-to=type=registry,ref=$(DOCKER_CACHE_TAG),mode=max \
		--push
	@docker pull $(DOCKER_PACKAGE_NAME)
endif
.PHONY: build
```

**Note**: `DOCKER_PACKAGE_NAME` & `DOCKER_CACHE_TAG` here are both docker image names. You also require to export env var `DOCKER_BUILDKIT=1`. In ci you do not have to set `PUSH`. In `cd` you set `PUSH` env var to `1` or `yes` to push all images (cache & casual one).

For example we manage that to package through docker a lambda. Produced artifact is already the same if no code changes in our ci/cd process, thanks to buildx and its cache mecanism.

image

sanmai:
That seems like a superior approach, ngl.

shouze:
@sanmai with docker 19.03 the buildx setup Makefile target I've posted can be ignored as buildx is built-in in [since docker 19.03 you just have to enable experimental cli support](https://docs.docker.com/buildx/working-with-buildx/#build-with-buildx).

Nyholm:
In the Github action community I read that it will be "some updates related to docker cache in min November". 

Im going to wait on that.

PR #98, opened 2020-11-23
Dont sleep, rather retry requests

Nyholm:
I will wait for new releases of Symfony and AsyncAws core to merge this


Nyholm:
I need the speed now!

PR #103, opened 2020-11-25
feat: add config to control built php versions

shouze:


Makefile
```diff
@@ -1,11 +1,11 @@
 SHELL := /bin/bash
-php_versions ?= 73 74
 layer ?= *
 parallel = $(if $(shell which parallel),true,false)
+resolve_php_versions = $(or $(php_versions),`jq -r '.php | join(" ")' ${1}/config.json`)
```
shouze, line 4: imply to have jq installed, should I change of something else? I can try doing it in php.
should look like that:
```sh
php -r 'echo join(" ", json_decode(file_get_contents("${1}/config.json"), true)["php"]);'
```

Nyholm: Yeah, But that imply that you have PHP installed =)

As long as it works on github actions then Im fine either way.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change log
 
+## 0.7.0
```
Nyholm, line 3: Lets
```suggestion
## 0.6.4
```

Makefile
Nyholm: Nice with `or` btw.

layers/amqp/Dockerfile
```diff
@@ -1,13 +1,13 @@
 ARG PHP_VERSION
-FROM bref/build-php-$PHP_VERSION:1.0.0-beta1 AS ext
+FROM bref/build-php-$PHP_VERSION:1.0.0 AS ext
```
Nyholm, line 2: Good!

Nyholm (approved):
Excellent work, thank you

Makefile
shouze: Yes, this is fully BC and very handy when you just want to rebuild some specific layer variant.

Nyholm:
I’ll merge this. Let’s see if it builds properly on master 👍 

Rock n’ Roll :)

PR #106, opened 2020-11-26
Prepare release for Bref 1.0

Nyholm:
Hm.. Only AMQP changed. We didn't get PHP8 here.. 

FYI @shouze


shouze:
Well, this sounds weird. BTW I've also opened https://github.com/brefphp/extra-php-extensions/pull/108.
Will investigate on the reason.

Nyholm:
This is super strange.. After the "build" step, we only have two layers: https://github.com/brefphp/extra-php-extensions/pull/106/checks?check_run_id=1457629075#step:6:10

Nyholm:
Ah, I found the bug =)

shouze:
Here we go, réajuster in time for php80 release

Issue #104, opened 2020-11-25
Imagick extension stopped working


chekalsky:
Couple of days ago imagick extension I've used stopped working. I thought it was connected to #99 but apparently it's not.

I've made very simple bref app to demonstrate it.
https://github.com/chekalsky/bref-imagick-issue

<img width="686" alt="image" src="https://user-images.githubusercontent.com/241584/100290443-a4f5ff00-2f7b-11eb-8aad-ab3eccc45678.png">

Error:
`PHP Warning:  PHP Startup: Unable to load dynamic library '/opt/bref-extra/imagick.so' (tried: /opt/bref-extra/imagick.so (libgomp.so.1: cannot open shared object file: No such file or directory), /opt/bref/lib/php/extensions/no-debug-non-zts-20190902//opt/bref-extra/imagick.so.so (/opt/bref/lib/php/extensions/no-debug-non-zts-20190902//opt/bref-extra/imagick.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0`

danniehansen:
Not sure if it's related. Also experiencing exactly the same with GD after upgrading to Bref 1.0 and added GD layer / ini.

jenschude:
https://github.com/brefphp/bref/blob/master/runtime/layers/tests.php

jenschude:
Btw there was once a test added to bref itself if extensions are loadable. This could help and prevent such issues in the future. ;)

See https://github.com/brefphp/bref/blob/master/runtime/layers/tests.php

hvipana:
> Not sure if it's related. Also experiencing exactly the same with GD after upgrading to Bref 1.0 and added GD layer / ini.

```
PHP Warning:  PHP Startup: gmp: Unable to initialize module
Module compiled with build ID=API20190902,TS
PHP    compiled with build ID=API20190902,NTS
These options need to match
 in Unknown on line 0
PHP Warning:  PHP Startup: Unable to load dynamic library '/opt/bref-extra/gd.so' (tried: /opt/bref-extra/gd.so (libpng12.so.0: cannot open shared object file: No such file or directory), /opt/bref/lib/php/extensions/no-debug-non-zts-20190902//opt/bref-extra/gd.so.so (/opt/bref/lib/php/extensions/no-debug-non-zts-20190902//opt/bref-extra/gd.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
```

chekalsky:
+ #109

[...]

PR #111, opened 2020-11-26
Test layers to make sure they work

### LINKED ISSUE #16: Automatically enable extensions?
Would it make sense to automatically enable extensions when the layer is used?

That would simplify the installation steps: just import the layer and that's it.

How? PHP currently looks for `*.ini` files in 2 places:

https://github.com/brefphp/bref/blob/c3ca836239e2a9d5445235599b2e932de267ddb7/runtime/base/php-74.Dockerfile#L76

I *think* this works (I kinda remember some problems with this months ago, not sure), and I *think* it would be possible for the extra layers to include a `pgsql.ini` file (for example) in `/opt/bref/etc/php/conf.d` that would load the extension.

WDYT?

Nyholm:
This PR moves CI to Github Actions because travis didnt want to run the jobs anymore.
It also runs a small PHP script for each extension to make sure it is loaded properly. I also run `php -m` and verify that the output does not contain any "Warning" or "Unable". 

I also make sure that most extensions are loaded automatically. 

This will fix #16 and maybe #104

It will also fix missing dependencies in the following extensions: 
- gd
- odbc-snowflake


Nyholm:
Oh, no.. Travis decided to hate me and not build this... something about negative balance..

Moving to GH actions it is...

PR #132, opened 2020-11-30
Feature/sqlsrv

atbreuer11:
Fix bug and add configuration file


layers/sqlsrv/Dockerfile
```diff
@@ -17,14 +17,17 @@ RUN pecl install -f pdo_sqlsrv
...
-  Driver      = ODBC Driver 17 for SQL Server\n\
-  Description = My ODBC Driver 17 for SQL Server\n\
-  Trace       = No'  > /tmp/odbc.ini
-RUN echo $'[ODBC Driver 17 for SQL Server]\n\
-  Description = Microsoft ODBC Driver 17 for SQL Server\n\
-  Driver      = /opt/microsoft/msodbcsql17/lib64/libmsodbcsql-17.6.so.1.1\n\
-  UsageCount  = 1'  > /tmp/odbcinst.ini
+Description = Microsoft ODBC Driver 17 for SQL Server\n\
+Driver      = /opt/microsoft/msodbcsql17/lib64/libmsodbcsql-17.6.so.1.1\n\
+UsageCount  = 1'  > /tmp/odbcinst.ini
```
Nyholm, line 30: Could you tell me why these lines are changed?

layers/sqlsrv/Dockerfile
```diff
@@ -17,14 +17,17 @@ RUN pecl install -f pdo_sqlsrv
 RUN cp /opt/bref/lib/php/extensions/no-debug-non-zts-*/sqlsrv.so /tmp/sqlsrv.so
 RUN cp /opt/bref/lib/php/extensions/no-debug-non-zts-*/pdo_sqlsrv.so /tmp/pdo_sqlsrv.so
 
+RUN echo $'extension=/opt/bref-extra/sqlsrv.so\n\
+extension=/opt/bref-extra/pdo_sqlsrv.so' > /tmp/ext.ini
```
Nyholm, line 21: Is there a problem to have both enabled?

Nyholm (commented):
Make sure to enable the test too. Just remove the first `exit(0)`.

layers/sqlsrv/Dockerfile
atbreuer11: This change resolved the following issue for me: `SQLSTATE[IM002]: [unixODBC][Driver Manager]Data source name not found, and no default driver specified (SQL: select * from [language])`. I suppose those configuration files are rather strict.

layers/sqlsrv/Dockerfile
atbreuer11: Should be fine. I have had them both enabled in the past with no issues.

atbreuer11:
Removed `exit(0)` and added a test for the PDO sqlsrv extension

layers/sqlsrv/Dockerfile
Nyholm: Thank you

Nyholm (approved):
Thank you

PR #143, opened 2020-12-13
Add php8 support for imagick

chekalsky:


layers/imagick/Dockerfile
```diff
@@ -35,9 +35,9 @@ RUN make install
 
 # Compile the php imagick extension
 WORKDIR ${IMAGICK_BUILD_DIR}
-RUN pecl download imagick
-RUN tar xzf imagick-3.4.4.tgz
-WORKDIR ${IMAGICK_BUILD_DIR}/imagick-3.4.4
+RUN git clone https://github.com/Imagick/imagick
+WORKDIR ${IMAGICK_BUILD_DIR}/imagick
+RUN git checkout 132a11fd26675db9eb9f0e9a3e2887c161875206
```
Nyholm, line 40: Hm.. This will install an unreleased version of Imagick. Is that safe?

Nyholm: I would be fine with this if we used a stable release on PHP7

chekalsky: Oh, there is quite tough convo about that in issues: https://github.com/Imagick/imagick/issues/358 so.. I don't know, tests looks fine.

Sury is also compiling it from master. Also: https://github.com/oerdnj/deb.sury.org/issues/1502

chekalsky: @Nyholm can I use `RUN if [[ "$PHP_VERSION" = "74" ]]` hack or there is neater way to do it?

Nyholm: That is something I had in mind, yes.

chekalsky: @Nyholm done

layers/imagick/Dockerfile
```diff
@@ -35,9 +36,11 @@ RUN make install
...
 # Compile the php imagick extension
 WORKDIR ${IMAGICK_BUILD_DIR}
-RUN pecl download imagick
-RUN tar xzf imagick-3.4.4.tgz
-WORKDIR ${IMAGICK_BUILD_DIR}/imagick-3.4.4
+# Temporary workaround for getting right imagick both for PHP 7.* and 8.*
+# @see https://github.com/brefphp/extra-php-extensions/pull/143/files#r541977286
+RUN if [[ "${PHP_VERSION}" = "73" ]] || [[ "${PHP_VERSION}" = "74" ]]; then pecl download imagick-3.4.4 && tar xzf imagick-3.4.4.tgz && mv imagick-3.4.4 imagick ; else git clone https://github.com/Imagick/imagick ; fi
+WORKDIR ${IMAGICK_BUILD_DIR}/imagick
+RUN if [[ "${PHP_VERSION}" != "73" ]] && [[ "${PHP_VERSION}" != "74" ]]; then git checkout 132a11fd26675db9eb9f0e9a3e2887c161875206 ; fi
```
Nyholm, line 43: Just some small things to reduce the complexity
```suggestion
RUN if [[ "${PHP_VERSION}" != "80" ]]; then pecl download imagick-3.4.4 && tar xzf imagick-3.4.4.tgz --directory ./imagick ; else git clone https://github.com/Imagick/imagick ; fi
WORKDIR ${IMAGICK_BUILD_DIR}/imagick
RUN if [[ "${PHP_VERSION}" = "80" ]]; then git checkout 132a11fd26675db9eb9f0e9a3e2887c161875206 ; fi
```

chekalsky: Although need to remember to support PHP 8.1 in case imagick will not provide a new version before new PHP version will be evailable :)

PR #163, opened 2021-01-20
Add msgpack PHP8 layer

Th3Mouk:


Th3Mouk:
Hi ! I'm currently using msgpack on PHP 8.0.1 but not yet on the lambda. I'm discovering layers, doing my own, with amqp + PHP8. ATM i'm not sure how i can test it easily.

Th3Mouk:
Just found [this documentation](https://github.com/brefphp/extra-php-extensions/blob/master/docs/create_your_own_extension_layer.md) i will use it to build and upload two php8 layers on our staging/preprod env

Th3Mouk:
@mnapoli this layer working, I got a phpunit test suite testing the result of `msgpack_*` on `74` and this layer. And tests are passing

Th3Mouk:
@mnapoli tested in pre-production too and it work. I have an `amqp` layer for PHP8, based on [this commit](https://github.com/php-amqp/php-amqp/commit/df1241852b359cf12c346beaa68de202257efdf1) instead of last release if you want too

mnapoli (approved):
Perfect, thanks for testing on Lambda.

@Nyholm not sure if there's anything else to check so I'll just approve and let you double check.

Nyholm:
Since our [small tests suite](https://github.com/brefphp/extra-php-extensions/blob/0.9.5/layers/msgpack/test.php) dont complain, I can only assume it is safe. Thank you @Th3Mouk for verifying manually too. 

About AMQP, I think we should wait on https://github.com/php-amqp/php-amqp/issues/386 or do some special thing like we do with Memcache.

PR #181, opened 2021-02-23
Updates & fix for sqlsrv extensions

### LINKED ISSUE #179: SQLSRV extension error with libmsodbcsql
Starting with release 0.9.6, getting an error with the SQLSRV layer when connecting with `sqlsrv_connect()`. Using Bref 1.2.0.

```
[{
	"SQLSTATE": "01000",
	"code": 0,
	"message": "[unixODBC][Driver Manager]Can't open lib '\/opt\/microsoft\/msodbcsql17\/lib64\/libmsodbcsql-17.6.so.1.1' : file not found"
}]
```

Downgrading to `bref/extra-php-extentions` 0.9.5 fixes it while still using Bref 1.2.0.

atbreuer11:
This should hopefully fix #179 which seems to be caused by an outdated reference to the driver file

It looks like the new driver file extension has the minor version removed, so hopefully this won't be an issue again
- https://github.com/atbreuer11/extra-php-extensions/blob/4386b3a33363004d23c361673ac3b06d779fbf93/layers/sqlsrv/Dockerfile#L25

I added specific versions for everything as an additional safeguard


atbreuer11:
@chrishanline I'm not sure why your test case from #179 is failing here. I wasn't able to get it to pass locally, though the layer is working for me. Any ideas?

chrishanline:
https://github.com/brefphp/extra-php-extensions/blob/2c5c5ba5ff44e18ab8943cc31582b7cce738e71d/layers/sqlsrv/Dockerfile#L29

https://github.com/brefphp/extra-php-extensions/blob/4386b3a33363004d23c361673ac3b06d779fbf93/layers/sqlsrv/Dockerfile#L25

@atbreuer11 are we sure the ODBC Driver 17 moved from `/opt/microsoft/msodbcsql17/lib64/` as previous to `/opt/bref/lib64/`? the failing test suggests that `ODBC Driver 17 for SQL Server` can't be found and I believe the odbcinst.ini file is what is used to determine the path.

atbreuer11:
@chrishanline Both paths work for me, and neither got the test to pass.

It appears that /usr/lib64/libmsodbcsql-17.so is symlinked to /opt/microsoft/msodbcsql17/lib64/libmsodbcsql-17.#.so.1.1, where the # changed from 6 to 7 (I suspect the root of the problem). Since both the symlink and the original file get copied, I figured to use the former inside odbcinst.ini as it doesn't include the minor version.

However I realized that I could just copy over the configuration files from the installation as the paths are the same. The path used by the installed odbcinst.ini is the /opt one, and it should handle the minor version automatically.

Nyholm:
I like that we specify the exact version of the extensions. Well done. 

Is there any way we can make the test pass? Or is the test wrong?

atbreuer11:
It looks like the unixODBC driver sourced from Microsoft was the culprit, replacing it with the version from amzn2-core/amzn-main gets the test to pass. I retested the extensions and both continue to work.

layers/sqlsrv/Dockerfile
```diff
@@ -3,32 +3,20 @@ FROM bref/build-php-$PHP_VERSION:1.1.4 AS ext
 
 ENV LD_LIBRARY_PATH=/usr/lib:/usr/lib64:$LD_LIBRARY_PATH
 
-RUN curl http://mirror.centos.org/centos/7/os/x86_64/Packages/unixODBC-2.3.1-14.el7.x86_64.rpm > /tmp/unixODBC-2.3.1-14.el7.x86_64.rpm
-RUN yum -y install /tmp/unixODBC-2.3.1-14.el7.x86_64.rpm
-RUN curl http://mirror.centos.org/centos/7/os/x86_64/Packages/unixODBC-devel-2.3.1-14.el7.x86_64.rpm > /tmp/unixODBC-devel-2.3.1-14.el7.x86_64.rpm
-RUN yum -y install /tmp/unixODBC-devel-2.3.1-14.el7.x86_64.rpm
+RUN yum -y install unixODBC-devel
```
Nyholm, line 6: Could we specify a version here?

atbreuer11: I actually did have one specified, but then I noticed that `amzn-main` and `amzn2-core` contain different versions of that driver.
Screen Shot 2021-02-28 at 12 50 52 PM

Nyholm: Oh, Okey. We ware not using amazonlinux:1 anymore, could we keep using 2.3.1-14?

atbreuer11: Sure, I reverted that change

Nyholm:
Wohoo!

Thank you for fixing this.

layers/sqlsrv/test.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+putenv('ODBCSYSINI=/opt/microsoft/conf/');
```
deleugpn, line 3: Do we need to define this variable on our Lambda function as well?

PR #183, opened 2021-02-26
[Imagick] Don't specify patch versions when copying files

Nyholm:
This will fix that master is red


layers/imagick/test.php
```diff
@@ -12,9 +12,6 @@
     }
 }
 
-exit(0);
-
-// This test is disabled because it causes Docker to segfault. See https://github.com/brefphp/extra-php-extensions/pull/156
```
Nyholm, line 17: The imagick extension had dependencies that updated. I fixed that and just for fun, I tried to remove these lines. And everything works =)

FYI @victormacko

victormacko: Nice one @Nyholm !
I just gave this a go, unzipped the layer, and it's contents has blown out from 56mb (ish) to roughly 218mb due to multiple copies of the libraries being brought across (eg. libtiff.so, libtiff.so.5, libtiff.so.5.2.0 in one case, libgomp.so.1 and libgomp.so.1.0.0 in another).
Were these files previously symlinked or do we need to only bring across one copy of each of the libraries somehow?

Nyholm: Thank you for checking. I could not deploy this because it was too large. 

I was sure these files was symlinked. I’ll revert this pr to only copy libgomp.so.1 and not libgomp.so.1.0.0

mnapoli: FYI I think symlinks are not supported in layers, we have the same problem in Bref layers where some files are duplicated like this. We could trim down the layer sizes if we managed to make it work. I tried once, but didn't get it to work.

Nyholm: I fixed the issue in https://github.com/brefphp/extra-php-extensions/pull/187 where I just copy each lib once by specifying the major version.

PR #185, opened 2021-03-01
Release version 0.10.2

Nyholm:


Nyholm:
Hm.. 

FYI @mnapoli, The Imagick layer is larger than 64MB... So we cannot publish it. I need to make it thinner somehow.

PR #156, opened 2021-01-05
Updated test for writing text to images using imagick 

victormacko:
As requested in issue #155 

This PR will fix #155

PR #186, opened 2021-03-02
Include latest ghostscript library in imagick extension

guillaumesmo:
The ImageMagick library is compiled with the GhostScript library. It uses the amazonlinux-provided version which is rather outdated (9.06 - almost 9 years old). This PR includes the compilation of the latest version (9.53.3) in the image. 

Also, the included ghostscript library has no built-in initialization scripts, so reading a PDF file (and probably also EPS files) with PHP imagick fails with the following error:
GPL Ghostscript 9.06: Can't find initialization file gs_init.ps.\n @ error/ghostscript-private.h/InvokeGhostscriptDelegate/178
this means the ghostscript support is useless within the extension now


layers/imagick/Dockerfile
```diff
@@ -90,6 +98,7 @@ COPY --from=ext /usr/lib64/libpaper.* /opt/bref/lib/
 COPY --from=ext /usr/lib64/libopenjp2.* /opt/bref/lib/
 COPY --from=ext /opt/bref/lib/libde265.* /opt/bref/lib/
 COPY --from=ext /opt/bref/lib/libheif.* /opt/bref/lib/
+COPY --from=ext /opt/bref/lib/libgs.* /opt/bref/lib/
```
Nyholm, line 101: Why include it down here?

Nyholm:
Thank you for fixing this. Could you also update the test to make sure this will always work?

Nyholm:
Also, do a small rebase. I updated Imagick in #187

guillaumesmo:
@Nyholm thank you for your review. I would like to discuss an alternative approach though

I see that --with-gslib=yes was added in the ImageMagick configure line last month, together with --with-freetype=yes, in a PR which fixes text rendering: #156. I wonder if gslib was really needed for that

The reason why I'm asking is that if using --with-freetype=no, this would save some disk space because of no dependency on libgs.so, and the user can ship any gs binary as an additional layer in the lambda function, eg https://github.com/shelfio/ghostscript-lambda-layer

Nyholm:
> I wonder if gslib was really needed for that

I dont know. 

> this would save some disk space

How much?

----------

I've honestly never used imagick and I know very little about what ghostscript is or what freetype is. That is why Im trying to push every new feature to add a test.

guillaumesmo:
gs is a binary/library which is used to convert postscript files (PDF, EPS) to image files. ImageMagick delegates conversion of those files to gs either through the library (--with-gslib=yes) or through the binary (--with-gslib=no)

gs depends on a number of files (postscript init scripts and ICC profiles) to be distributed with the library. By default those files are compiled into the .so file, but that was either disabled (--disable-compile-inits) or not yet supported in the version shipped with al2, which is why gs doesn't work currently in the bref imagick extensions (only the .so file is copied) as indicated in the error "Can't find initialization file gs_init.ps"

> > this would save some disk space
> 
> How much?

the most obvious one is the removal of libgs.so.9 itself which saves 16MB
but I believe some other libraries were added as dependencies of gs and should not be needed either
eg all the X libraries (because al2 gs was not built --without-x):
1.3M libX11.so.6
16K libXau.so.6
75K libXext.so.6
419K libXt.so.6
162K libxcb.so.1

guillaumesmo:
it seems I was wrong about the version shipped with al2, it's 9.25 and not 9.06 like the error message suggested. That doesn't change anything to the problem though.

Also, my initial commit was not working as expected, I have added a test which fails in the current master (and in my initial PR)

I have tried many different ways to get the library working but I always have the following issue:
"Unrecoverable error: ioerror in setdistillerparams"
I have no idea what the error stands for, and googling it doesn't is not really helpful. 

```
    "errorMessage": "PostscriptDelegateFailed `[ghostscript library 9.53] -sstdout=%stderr -dQUIET -dSAFER -dBATCH -dNOPAUSE -dNOPROMPT -dMaxBitmap=500000000 -dAlignToPixels=0 -dGridFitTT=2 '-sDEVICE=pngalpha' -dTextAlphaBits=4 -dGraphicsAlphaBits=4 '-r72x72'  '-sOutputFile=/tmp/magick-c_rn-YoHAYTEtjJEVGuax9gR9jb2HWaD%d' '-f/tmp/magick-XYF9Yv5xc8fv3ABqiIs88FIThMrp8ybw' '-f/tmp/magick-2rBfIf8nyMVrVyvpb6nT2-oU-xNfYfcp'': Unrecoverable error: ioerror in setdistillerparams\nOperand stack:\n    --nostringval--\n @ error/ghostscript-private.h/InvokeGhostscriptDelegate/178",
```

also I wonder why it calls gs with 3 different input files, while it's just one pdf.

The following issue seems to be related, where the suggestion is to not use libgs:
https://github.com/ImageMagick/ImageMagick/issues/1699#issuecomment-533033077

---

However, it does work by using gs instead of libgs, and as you suggested I have included it in the layer.

Nyholm:
Excellent. Great work @guillaumesmo

PR #187, opened 2021-03-03
Specify versions

Nyholm:
The symlinks does not work properly. So we copy the same file twice which means the layer is too large for lambda.


layers/imagick/Dockerfile
```diff
@@ -59,37 +59,42 @@ COPY --from=ext /opt/bref/lib/libMagickWand-7.Q16HDRI.so.8.0.0 /opt/bref/lib/lib
...
+COPY --from=ext /usr/lib64/libidn.so.11 /opt/bref/lib/libidn.so.11
+COPY --from=ext /usr/lib64/libpaper.so.1 /opt/bref/lib/libpaper.so.1
+COPY --from=ext /usr/lib64/libopenjp2.so.7 /opt/bref/lib/libopenjp2.so.7
+
+COPY --from=ext /opt/bref/lib/libde265.so.0 /opt/bref/lib/libde265.so.0
+COPY --from=ext /opt/bref/lib/libde265.a /opt/bref/lib/libde265.a
+COPY --from=ext /opt/bref/lib/libde265.la /opt/bref/lib/libde265.la
+COPY --from=ext /opt/bref/lib/libheif.so.1 /opt/bref/lib/libheif.so.1
+COPY --from=ext /opt/bref/lib/libheif.a /opt/bref/lib/libheif.a
+COPY --from=ext /opt/bref/lib/libheif.la /opt/bref/lib/libheif.la
```
Nyholm, line 97: @victormacko

I needed these *.a and *.la files. That is why the build is green.

Nyholm:
The zipped layer is now 29MB. 

It was 19MB, after #183 it got up to 54MB.

PR #201, opened 2021-03-23
Just fixed some errors and unclear phrasing.

djzevenbergen:
I'm still a little unsure of what  Line 48: `If using a KMS it
will make it easy to globally change across numerous Lambda's the location of
a daemon, the log level or the license key if you need to change where the data
reports.` means. It seems like it's saying "If you're using a KMS and you set up your configuration to use ENVIRONMENT VARIABLES, it will be easier to globally change the location of the daemon, the log level, or the license key (if you need to change where the data reports) across multiple Lambdas.  

I'm not a SME so I'm not 100% sure what it is supposed to say, but it should maybe be clarified.

I work at New Relic - DJ Zevenbergen

### DIFF
--- a/docs/newrelic.md
+++ b/docs/newrelic.md
@@ -6,12 +6,12 @@ Install the PHP extension like any other extension. When you added the layer, yo
 need add some extra php configuration for you License Key and Daemon Address.
 The License Key should be your New Relic account License Key to ensure this app
 reports to your account. The daemon address should be the location of a remote New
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks a lot, that's definitely an improvement!

Regarding your question, I'm not 100% what the original author meant, but I think it's more: if you have multiple lambdas, you can put the values in KMS. Then if you change them in KMS, they will be changed for all lambdas referencing them.

PR #203, opened 2021-03-26
Support PHP 8.0

### LINKED ISSUE #202: PHP 8.0 layers
Hello,

I've just tried to use AMQP for PHP 8.0 and the layer does not exist. How could I help please ? :)

https://exploit.cz/how-to-compile-amqp-extension-for-php-8-0-via-multistage-dockerfile/

https://github.com/php-amqp/php-amqp/issues/386#issuecomment-795115965

vasilvestre:
This will fix #202


layers/amqp/Dockerfile
```diff
@@ -1,5 +1,6 @@
 ARG PHP_VERSION
 FROM bref/build-php-$PHP_VERSION:1.1.4 AS ext
+ARG PHP_VERSION
```
Nyholm, line 3: You were missing this line. 

Since this is a "multi image build", we technically start a new image after "FROM", that is why we need to redefine the ARG.

Nyholm:
Thank you for this PR. I added one small commit and things seams to build perfectly.

layers/amqp/Dockerfile
vasilvestre: Thank you a lot ! I didn't knew it :)

PR #224, opened 2021-05-09
Upgrade NewRelic Agent (with PHP8 support)

starred-gijs:


starred-gijs:
Thanks for merging! Just so that I understand, this package needs a new release to make the new layer available?
Now I get `Unknown Bref extra layer named "newrelic-php-80"`

Nyholm:
Correct. Let's prepare a release now

PR #231, opened 2021-05-25
Add layer for Symfony Runtime

Nyholm:
This is not a PHP extension, but it is a "bref extra". 

Using this layer together with the `runtime/bref` package, will make Symfony 5.3 application to run with ease. 

See more at https://github.com/php-runtime/bref


mnapoli:
👍  sounds good (per https://github.com/brefphp/bref/pull/889#issuecomment-847846006)

Nyholm:
Lovely. 

Thank you for the review.

PR #256, opened 2021-07-26
Add rdkafka extension

chrisnharvey:
Added the [php-rdkafka](https://github.com/arnaud-lb/php-rdkafka) extension.

Lambda functions can be triggered by Kafka events using MSK in the Serverless framework. [See here](https://www.serverless.com/framework/docs/providers/aws/events/msk/).

I've added this extension so we can not only consume Kafka events, but produce them also which is perfect for event driven architecture that Bref is very well suited for.


Nyholm (approved):
Great idea. Thanks

PR #265, opened 2021-09-01
Updating Oracle Instant Client version and lib path

mixaster:
Verified this works in a running Lambda, connected to Oracle 19c


Nyholm:
Thank you. 

Do we need to update the test for this? Ie to make sure the updated version is working?

Nyholm:
Okej. Cool cool. Thank you for checking. 

If we can do a version check or something. Feel free to follow up this PR with that test.

PR #272, opened 2021-09-25
Add support for Tidy extension

healthierjacobsen:


Nyholm:
Thank you for this PR. 

The build fails. It seams like you need to move `libtidy-0.99.so.0` to the final image too.

healthierjacobsen:
> Thank you for this PR.
> 
> The build fails. It seams like you need to move `libtidy-0.99.so.0` to the final image too.

@Nyholm I tried adding that in but am still failing. Any other advice?

Nyholm:
That did make a difference. Look at the last 20 lines of the output. 

Now it says:
> undefined symbol: zval_used_for_init

healthierjacobsen:
> Awesome. Build is green. Are you happy with this now?

Yes, it is finally green! I feel like I understand all of this much better now. Thank you for your assistance.

PR #280, opened 2021-10-29
New Extension Layer: Redis with Igbinary Serializer

sbennett-ihasco:
Added PHP 7.4 / 8.0 Redis extension compiled to use Igbinary serializer


Nyholm (approved):
Awesome. Thank you

Issue #281, opened 2021-10-29
Redis layer missing on PHP 8.1


aran112000:
The Redis layer which was added by @shouze under the following commit for PHP 8.1 doesn't appear to actually work:
https://github.com/brefphp/extra-php-extensions/commit/f5fe4498b58d662c2725896373ea5ee6de015564

When registered for a function PHP reports the Redis function never loaded - verified with `get_loaded_extensions()`. No sign of any startup errors from PHP which I would have expected, any idea how to debug/resolve this?

sbennett-ihasco:
I noticed the same thing building a redis+igbinary extension for 8.1. Staying with 8.0 for now.

aran112000:
@sbennett-ihasco we had other issues with 8.0 as the Memcached extension is missing - I've seen there's a MR sorting that out awaiting the next release

aran112000:
Managed to get the startup errors outputting which confirms both Redis and Memcached were compiled against the wrong PHP version:
```
PHP Warning:  PHP Startup: memcached: Unable to initialize module
Module compiled with module API=20201009
PHP    compiled with module API=20210902

These options need to match in Unknown on line 0
```
```
PHP Warning:  PHP Startup: redis: Unable to initialize module
Module compiled with module API=20201009
PHP    compiled with module API=20210902

These options need to match in Unknown on line 0
```

aran112000:
closed by #282

PR #282, opened 2021-10-30
Update Redis Dockerfile to fix wrong version compiled

aran112000:
After discovering the Redis extension was failing on PHP startup for 8.1 (https://github.com/brefphp/extra-php-extensions/issues/281), I've now tracked down the extension was being compiled against PHP 8.1 **BETA** 3, rather than RC4. Removing the version on the Docker image should help to keep it in sync going forward with the PHP versions of future Bref releases.

### DIFF
--- a/layers/redis/Dockerfile
+++ b/layers/redis/Dockerfile
@@ -1,5 +1,5 @@
 ARG PHP_VERSION
-FROM bref/build-php-$PHP_VERSION AS ext
+FROM bref/build-php-$PHP_VERSION:1.3.3 AS ext
 
 RUN pecl install --force redis
 RUN cp `php-config --extension-dir`/redis.so /tmp/redis.so
--- a/layers/redis/Dockerfile
+++ b/layers/redis/Dockerfile
@@ -1,5 +1,5 @@
 ARG PHP_VERSION
-FROM bref/build-php-$PHP_VERSION:1.2.13 AS ext
+FROM bref/build-php-$PHP_VERSION AS ext
 
 RUN pecl install --force redis
 RUN cp `php-config --extension-dir`/redis.so /tmp/redis.so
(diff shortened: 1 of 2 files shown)


aran112000:
Once this change has been reviewed/accepted, I'll do the same for the Memcached one too which is also failing currently

aran112000:
@deleugpn You're far more knowledgeable than me when trying to get everything compiling nicely. I believe the tests are being run against old PHP versions too, which is why they were passing before (when they were actually failing when deployed to AWS), now the Lambda layer versions match, the tests are failing. What's considered the *right* way to handle this, please?

aran112000:
Morning @Nyholm; any chance you'd have a chance this week to review/merge this change so we can carry on with our testing, please? Would get great to get a release with this and the Memcached changes from last week tagged so the layers go up :)

Nyholm:
Hey. 
This looks good. 

Could you follow up this PR with the same change on all the layers?

aran112000:
> Could you follow up this PR with the same change on all the layers?

Sure thing, out of interest is there are reason versions are explicitly specified, rather than just using the latest version of the base images?

Nyholm:
It is generally a bad practice to use `latest` since we dont know if latest is `master`, "the highest version" or "the next version". It is just the latest build. That is also no support for semver. 

We could do `:1`, but at the time I dont think Bref tagged major version only. Do they do that now?

So the short answer is: `latest` is bad, hence we fixed the version.

aran112000:
> So the short answer is: `latest` is bad, hence we fixed the version.

Makes total sense, I've pushed out a new PR updating the version on all other layers now, I did catch two that were using latest already which I've updated to use 1.3.3 instead based on this.

Thanks for your help and the quick release

PR #284, opened 2021-11-01
Updated the base Bref PHP layer versions

aran112000:
Updates the base Bref PHP docker image version to match test cases


Nyholm:
Thank you

PR #286, opened 2021-11-01
Temp fix for deployment issues

robchett:
Looking at https://github.com/brefphp/extra-php-extensions/actions/workflows/blackfire.yml it seems the deploy process broke 3 days ago and the only changes to the deploy process in that time were updates to symfony/http-client from 5.3.8 to 5.3.10


Nyholm:
Thank you for this. 
You are correct. I just did the same fix locally to release the next version.

PR #290, opened 2021-11-22
Bumping AMQP and Blackfire to allow support for PHP 8.1

aran112000:


Nyholm:
Thank you. 

It looks like some layers are not as easily upgraded to 8.1 

Check the CI, whatever layer is red, just revert that config file. Then you can follow up this PR with fixes for each layer

sbennett-ihasco:
To save you time, Redis+Igbinary has an 8.1 fix. :-)

https://github.com/brefphp/extra-php-extensions/pull/295

Nyholm:
I've removed the layers that failed.

PR #293, opened 2021-11-24
Install sqlsrv on PHP 8.1

ddeboer:


layers/blackfire/config.json
```diff
@@ -2,6 +2,7 @@
     "php": [
         "73",
         "74",
-        "80"
+        "80",
+        "81"
```
GrahamCampbell, line 6: This change is not related, and should be in a different PR

GrahamCampbell (changes requested):

layers/blackfire/config.json
ddeboer: Oops, sorry, that shouldn’t be there. Removed.

Nyholm:
Thank you. 

Could we do a workaround so the beta release is not installed on php 8.0 and lower?

ddeboer:
@Nyholm Done!

Nyholm (approved):
Thank you

PR #295, opened 2021-11-26
Redis with Igbinary Serializer: PHP 8.1

sbennett-ihasco:
Updated bref build image version from 1.2.13 to 1.3.3
Added 81 to config.json
Tested locally and as AWS layers in a bref deployment


Nyholm:
Great. Thank you. 

Could you help me understand what the two removed arguments means?

> enable-redis-lzf="no" enable-redis-zstd="no"'

sbennett-ihasco:
> Could you help me understand what the two removed arguments means?
> 
> > enable-redis-lzf="no" enable-redis-zstd="no"'

My thinking is that explicitly disabling Redis LZF compression in the layer is redundant as it's disabled by default (based on the default using the manual PECL installer). Enabling it would be adding something other than Igbinary support to the layer. However, the LZF algorithm does appear to make for faster compression (https://doordash.engineering/2019/01/02/speeding-up-redis-with-compression/).

It's the same for ZTSD compression (which enables a greater compression ratio): https://github.com/chadnickbok/redis-zstd-module

Thoughts on explicitly enabling these?

Nyholm:
Ah, okey. So the default behavior is not ti install lzf and zstd. Correct?
So we just removed parameters no doing anything. 

Is so, Im fine =)

sbennett-ihasco:
> So the default behavior is not ti install lzf and zstd. Correct? So we just removed parameters no doing anything.

I believe so, but I'm now going to check with a default Redis PECL install... :-D

sbennett-ihasco:
The default PECL Redis install is the same. 

Explicitly setting `enable-redis-lzf="yes"` results in this:

image

Explicitly setting `enable-redis-zstd="yes"` requires `libzstd`.

If anyone wants these options, I think they should be separate layers (could get busy round here!). 

We're good to go.

Nyholm:
Awesome. Thank you

PR #302, opened 2021-12-07
Add RdKafka layer to readme

rdarcy1:
Added to readme with version 7.4 to keep it consistent. I wonder if it would be better to change the readme to use "PHP *x*.*y*" and then reference `${bref-extra:rdkafka-php-xy}`?


Nyholm:
Thank you. 

Im not sure if it's matter. Currently it is a clear example and people understand `7.4` better than `x.y`.

PR #304, opened 2021-12-20
Bump libde265, libheif versions 

takeru56:
Hello. I’m a master’s student, and investigating whether updates from one project are useful for another project.
In this pull request, I am updating libde265 from 1.0.5 to 1.0.8 and libheif from 1.62 to 1.12.0. 
Since these two updates are being done in　https://github.com/carsales/pyheif/commit/06ca457f872045144e139b4117af7c99ab4f2ce2, I’m wondering if this project can update the libraries as well.

### DIFF
--- a/layers/imagick/Dockerfile
+++ b/layers/imagick/Dockerfile
@@ -9,18 +9,18 @@ RUN LD_LIBRARY_PATH= yum -y install libwebp-devel wget libpng-devel libjpeg-deve
 
 # Compile libde265 (libheif dependency)
 WORKDIR ${IMAGICK_BUILD_DIR}
-RUN wget https://github.com/strukturag/libde265/releases/download/v1.0.5/libde265-1.0.5.tar.gz -O libde265.tar.gz
+RUN wget https://github.com/strukturag/libde265/releases/download/v1.0.8/libde265-1.0.8.tar.gz -O libde265.tar.gz
 RUN tar xzf libde265.tar.gz
-WORKDIR ${IMAGICK_BUILD_DIR}/libde265-1.0.5
+WORKDIR ${IMAGICK_BUILD_DIR}/libde265-1.0.8
 RUN ./configure --prefix ${INSTALL_DIR} --exec-prefix ${INSTALL_DIR}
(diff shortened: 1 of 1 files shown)


Nyholm:
Hey. 

Lovely. What would the benefits for imagick be when we update these dependencies?

takeru56:
Imagick requires libheif to support HEIC format images, and libheif requires libde265 for image codecs. Although this dependency update is mainly a major update (support for AVIF format), I believe that keeping the dependencies up-to-date is important from a maintenance perspective. Version compatibility between libheif and libde265 is probably guaranteed since they are used in https://github.com/carsales/pyheif/commit/06ca457f872045144e139b4117af7c99ab4f2ce2.

GrahamCampbell (approved):

Nyholm:
Thank you. 

> I believe that keeping the dependencies up-to-date is important from a maintenance perspective. 

True, but so is stable software. Each change, big or small, is a risk for bugs and issues. If a change does not bring any value it should not be considered. 

> Version compatibility between libheif and libde265 is probably guaranteed... 

"probably guaranteed" does not sound like "guaranteed" =)

-----------

Note, Im not against this change. Im just curious why this is needed? How will it help you when you run imagick on with Bref? Will it introduce a new feature? Will it fix a bug for you? Or will it help you in any other way? 

Please elaborate so I understand the motivation behind this PR.

takeru56:
I thought it would be a good idea to update these libraries, not for any special features, but because there were some bug fixes (e.g., https://github.com/strukturag/libheif/commit/56c8a2613370562fc330af2c70c1510aa5fd9ff6, https://github.com/strukturag/libheif/commit/f454c27c1fc2562975e368d8822eeb6dd1ad1fab). Also, libheif@v1.6.2 seems to have a CVE-2020-23109 problem. However, I found that the proposed libde265@1.0.8 has a CVE-2021-35452 problem. So, I am not sure if this PR update is appropriate.

mnapoli:
@GrahamCampbell would these updates be superseded by #434 ?

GrahamCampbell:
Yes, this can be closed.

PR #321, opened 2022-02-09
OCI8: Update InstantClient and Add Support for PHP 8.1

mixaster:
Note: Not sure what the protocol is for the `build-php-` version, but I had to change it from 1.3.3 to 1.5.3 to get pecl to work with 8.1. Prior to that change, I was getting this error:

`pecl/oci8 requires PHP (version >= 8.1.0), installed version is 8.1.0RC4`


Nyholm (approved):
Great work. 

Thank you very much

PR #337, opened 2022-04-25
Adds ScoutAPM extension

asgrim:
Adds the `scoutapm` PECL extension - http://pecl.php.net/package/scoutapm


Nyholm (approved):
Great. Thank you

PR #344, opened 2022-05-19
update odbc snowflake driver and massively reduce layer size

Nemo64:
The odbc-snowflake layer was about 200mb in size.
Most of it was in the libSnowflake.so with 177mb.
I figured that it contains debug flags, that can be stripped.
This reduces the size of that so file to 73 mb.
Still big, but more than 100mb less.


Nyholm:
Wow, Thank you @Nemo64. This is awesome. 

> I figured that it contains debug flags, that can be stripped.

I appreciate that you added tests. But apart from that I have no way to verify this PR. I'll merge and let you try with Docker. =)

PR #355, opened 2022-08-13
Update to latest sqlsrv

GeNyaa:
The versions used are very outdated, this way we're using the latest version with every new build.


Nyholm:
I do like to lock the versions. That means no surprises when rebuilding. Could you please update to the latest versions?

GeNyaa:
@Nyholm Changed to latest sqlsrv version at this current time 5.10.1

Nyholm:
AWesoem!

PR #358, opened 2022-08-25
Add Elastic Application Performance Monitoring (APM) extension

mattrenner:
[Elastic APM](https://www.elastic.co/observability/application-performance-monitoring) is a performance monitoring
solution that has an agent / SDK for instrumenting multiple different languages, including PHP.

We've been using this for a few months successfully in a production environment so figure we should share the work that was done to allow adding this to Bref.

This PR adds the elastic-apm extension, along with some brief documentation on how to configure it with the experimental Lambda Extension also provided by Elastic.


Nyholm:
Lovely. Thank you

PR #374, opened 2022-11-19
Added PHP OpenSwoole Extension

lotharthesavior:
Hello,

I added here an [OpenSwoole](https://openswoole.com/) layer. I hope this help others. Please let me know if I missed something.

All tests passed.

Cheers,
Savio Resende


Nyholm:
Awesome. Thank you

mnapoli:
@lotharthesavior it seems like on master the build failed: https://github.com/brefphp/extra-php-extensions/actions/runs/3673601936/jobs/6210867538

Any idea?

lotharthesavior:
@mnapoli I responded to the other message about an issue that might be the same. I'll get back to you on that later today or tomorrow. After a quick look it seems that it is because I didn't  prepare support for php8.2, only for php8.1 and older. If so, I'll fix it as soon as I get to it.

mnapoli:
Oh that makes sense now! Thanks!

Nyholm:
FYI, I did a follow up in #379

@lotharthesavior if you find something, just add a new PR with the fix.

lotharthesavior:
I found that this version is not ready on some linux package managers (the ones I tested, specially the one used by lambda) doesn't have openswoole, and compiling it didn't help. I'll keep investigating and will create another PR as soon as I have the solution.

Here is the current PR: https://github.com/brefphp/extra-php-extensions/pull/380

@Nyholm @mnapoli

PR #380, opened 2022-12-16
OpenSwoole Version Fix

lotharthesavior:
This is a temporary fix until openswoole extension support for php8.2 is ready.


Nyholm:
Thank you

PR #381, opened 2022-12-19
Support PHP 8.2

### LINKED ISSUE #378: Support PHP 8.2
Creating a separate issue to track this, because in #375 it was decided to provide PHP 8.2 support for Bref 1.x before adding Bref 2 support.

Work was done in #376 to enable PHP 8.2. Can we just add `82` to the `config.json` files or is there more to be done?

ddeboer:
Fix #378


mnapoli:
It seems like newrelic and openswoole are failing. New Relic doesn't seem to support PHP 8.2 for example:

> unsupported version '8.2.0' of PHP found

Maybe these 2 extensions should be ignored for now?

starred-gijs:
Yes please ignore NewRelic for PHP 8.2, they dont support it yet.

Nyholm:
Thank you

PR #384, opened 2022-12-21
Release 0.12.2

Nyholm:


ddeboer:
> The security token included in the request is invalid.

Would it be so simple as to update the token?

mnapoli:
> eu-south-2

Maybe it's a matter of enabling this region in the AWS account?

GrahamCampbell:
Yeh. All new regions are opt-in since Hong Kong in 2019.

Nyholm:
Thank you. 
Enabling regions now

Nyholm:
Finally =)

PR #377, opened 2022-12-10
Extensions V2, speedier to distribute than V1 thanks to Bref V2 decisions :zap: 

shouze:
Closes #375

# TODO

- [ ] AMQP
- [ ] Blackfire
- [x] Calendar
- [x] Cassandra
- [x] Decimal
- [x] DS
- [ ] Elastic APM
- [ ] GD
- [x] gnupg
- [x] GMP
- [x] gRPC
- [ ] ~Igbinary~ ping @mnapoli @Nyholm already included into redis & memcached
- [ ] Imagick
- [x] IMAP
- [x] LDAP
- [x] Mailparse
- [ ] MaxMind DB
- [x] Memcache
- [x] Memcached
- [x] MongoDB
- [ ] ~MsgPack~ ping @mnapoli @Nyholm already included into redis & memcached
- [ ] Newrelic
- [ ] ODBC Snowflake
- [ ] Oracle
- [x] Pcov
- [ ] PostgreSQL
- [x] RdKafka
- [x] Redis
- [ ] ~Redis-Igbinary~ ping @mnapoli @Nyholm already included into redis & memcached
- [ ] Scout APM
- [ ] Scrypt
- [ ] SPX
- [x] SSH2
- [ ] Symfony Runtime
- [ ] Microsoft SQLSRV
- [x] Tidy
- [x] UUID
- [x] Xdebug
- [ ] Xlswriter
- [x] Yaml

PR #385, opened 2022-12-21
Add the `relay` extension

mnapoli:
This PR is based on #377, only the last commit is interesting. Once #377 is merged I can rebase this.

This PR adds the Relay (https://relay.so/) extension. It only adds it for Bref v2.

cc @tillkruss


layers/relay/test.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+if (!class_exists($class = \Relay\Relay::class)) {
```
mnapoli, line 3: @tillkruss Sounds like a reasonable way to test the extension works, right?

layers/relay/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# Install system libraries needed by Relay
+RUN yum install -y openssl11-libs # libssl 1.1 and libcrypto 1.1 (system version is 1.0)
+
+RUN <<'END' bash -e
+    mkdir -p /tmp/relay
+    export version=$(php -r "echo PHP_MAJOR_VERSION.'.'.PHP_MINOR_VERSION;")
+    curl -sSL "https://cachewerk.s3.amazonaws.com/relay/v$RELAY_VERSION/relay-v$RELAY_VERSION-php$version-centos7-x86-64.tar.gz" | tar -xz --strip-components=1 -C /tmp/relay
+    sed -i "s/00000000-0000-0000-0000-000000000000/$(cat /proc/sys/kernel/random/uuid)/" /tmp/relay/relay-pkg.so
+    cp /tmp/relay/relay-pkg.so /tmp/relay.so
+    cp /tmp/relay/relay.ini /tmp/ext-relay.ini
```
mnapoli, line 25: @tillkruss at the moment I copy the default relay.ini. But the docs recommends setting these keys:

```ini
relay.maxmemory = 32M
relay.eviction_policy = noeviction
relay.environment = production
relay.key = 1L0O-KF0R-W4RDT0-Y0URR3P-0RTMRBR-OCC0L1
```

I could pre-set these values except the key, but it _seems_ like these are already the defaults, correct? Any recommendation for Lambda? Or any particularity about Relay I should look into for Lambda (if you think of any)?

tillkruss: Correct, the default config is great, unless someone uses Relay heavily.

layers/relay/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN echo 'extension=igbinary.so' > /tmp/ext-igbinary.ini
+RUN echo 'extension=msgpack.so' > /tmp/ext-msgpack.ini
+
+# Install system libraries needed by Relay
+RUN yum install -y openssl11-libs # libssl 1.1 and libcrypto 1.1 (system version is 1.0)
+
+RUN <<'END' bash -e
+    mkdir -p /tmp/relay
+    export version=$(php -r "echo PHP_MAJOR_VERSION.'.'.PHP_MINOR_VERSION;")
+    curl -sSL "https://cachewerk.s3.amazonaws.com/relay/v$RELAY_VERSION/relay-v$RELAY_VERSION-php$version-centos7-x86-64.tar.gz" | tar -xz --strip-components=1 -C /tmp/relay
```
tillkruss, line 22:
```suggestion
   curl -sSL "https://builds.r2.relay.so/v$RELAY_VERSION/relay-v$RELAY_VERSION-php$version-centos7-x86-64.tar.gz" | tar -xz --strip-components=1 -C
```

tillkruss: We just switched over to R2, all future releases will live there. S3 is too expensive at scale.

mnapoli:
I opened #394 to support Bref v1.

mnapoli:
Superseeded by #394 and #406.

PR #390, opened 2023-01-10
Clarify that the redis extension is phpredis

mnapoli:
There are many options to use Redis with PHP ([phpredis](https://github.com/phpredis/phpredis), predis, etc.), it's confusing. I looked at the README and it wasn't clear to me whether it was the phpredis extension (required by a framework), so I thought we might as well clarify.


Nyholm:
Thank you

PR #397, opened 2023-02-02
Minor relay formatting

tillkruss:
`cd /tmp/relay` isn't needed


mnapoli (approved):
Woops sorry for the delay!

PR #406, opened 2023-03-04
Support Bref v2

mnapoli:


layers/amqp/Dockerfile
```diff
@@ -19,23 +20,16 @@ RUN set -xe; \
...
-COPY --from=ext /opt/bref/include/amqp_ssl_socket.h /opt/bref/include/amqp_ssl_socket.h
-COPY --from=ext /opt/bref/include/amqp_tcp_socket.h /opt/bref/include/amqp_tcp_socket.h
-
-COPY --from=ext /opt/bref/lib64/librabbitmq.a /opt/bref/lib64/librabbitmq.a
-COPY --from=ext /opt/bref/lib64/librabbitmq.so /opt/bref/lib64/librabbitmq.so
-COPY --from=ext /opt/bref/lib64/librabbitmq.so.4 /opt/bref/lib64/librabbitmq.so.4
-COPY --from=ext /opt/bref/lib64/librabbitmq.so.4.3.0 /opt/bref/lib64/librabbitmq.so.4.3.0
-COPY --from=ext /opt/bref/lib64/pkgconfig/librabbitmq.pc /opt/bref/lib64/pkgconfig/librabbitmq.pc
-
-COPY --from=ext /tmp/amqp.so /opt/bref-extra/amqp.so
```
Nyholm, line 40: Interesting. Can you elaborate on why it is save to remove these?

layers/cassandra/config.json
```diff
@@ -1,5 +1,4 @@
 {
     "php": [
-        "74"
```
Nyholm, line 3: Shouldn't we have at least one valid PHP version?

Nyholm (commented):
Thank you for your work.

layers/amqp/Dockerfile
Nyholm: https://github.com/brefphp/aws-lambda-layers/blob/main/utils/lib-copy/copy-dependencies.php

it seams scary =)

Should we just trust the tests?

mnapoli: Yeah that's the new script that automates copying the system dependencies used by a binary/library. The goal is to make all the Dockerfiles much simpler, easier to write, and more robust, so that we don't have to manually identify and pick the correct files.

It is definitely scary 😅 Any regression we can fix and add more tests to cover, but yeah, we are still at the risk of regressions for unexpected reasons.

All the `.so` files should be correctly copied by the script (that's what the script focuses on right now).

I don't know why some `.h`, .`pc` and `.a` files were copied in the exported layers. I know `.so` are libraries and we need them, but I don't know if those other files are actually needed at runtime.

Any idea? Maybe @shouze has an idea about these files?

layers/cassandra/config.json
mnapoli: I added PHP 8.0 for cassandra and scrypt.

scrypt compiled ok, but cassandra failed:

> pecl/cassandra requires PHP (version >= 7.2.0, version <= 7.99.99), installed version is 8.0.27

I think we shouldn't block this PR for a feature that didn't exist previously, I'll revert PHP 8 support for Cassandra and just leave it as is 🤷 Maybe someone else can contribute PHP 8 support later.

layers/amqp/Dockerfile
shouze: for `.h` and `.pc` files I see no reasons as they're files only useful for compilation/linking use cases.

`.a` files are static version of libraries so effectively I see no better reason to copy them because they should be linked inside a binary to work.

`.so` which are dynamic libraries should be enough.

mnapoli: Awesome, thanks for confirming!

Nyholm: Let's try this then =)

PR #408, opened 2023-03-10
Added parallel extension

hschimpf:
## Extension reference
- https://php.net/parallel

## Dependencies
- AWS Lambda PHP layer compiled with ZTS flag enabled

### Current `CI / Build and test layer` fails because this extension needs PHP with ZTS enabled
Please, refer to brefphp/aws-lambda-layers#63 for more details.


Nyholm:
Thank you. I put this on hold until there is a decision about ZTS.

FYI, I think I suggested to remove it years ago because "nobody is using pthreads or parallel anyways".

mnapoli:
👍 will post an update here when the ZTS PR is merged.

Basically what we are going with is: allow users to enable ZTS and recompile PHP runtimes, but it won't be in official Bref runtimes. So that extension will only work for those who will recompile PHP layers themselves.

Latest update https://github.com/brefphp/aws-lambda-layers/pull/63#issuecomment-1484772004

hschimpf:
exactly, it will be for special cases and will have to be compiled manually compiled along with the PHP runtimes with ZTS enabled.

hschimpf:
> FYI, I think I suggested to remove it years ago because "nobody is using pthreads or parallel anyways".

Parallelism programming is a completely different world on programming, and even more in PHP since is not a built-in feature of the language. I must be one of the few who use parallelism in PHP. Sharing data between threads and maintain synchronization between them requires a complex understanding of the libraries and your needs.

I had a special requirement on a project with a lot of data _(details on brefphp/aws-lambda-layers#63)_, and I was looking a way to reduce costs on AWS Lambda. Since lambda gives us with some cores _(based on memory settings)_, I pointed my goal in that direction and build some heavy demanding processes using parallelism. Another limitation on Lambda is the max execution time of 15min, so the more information I can process in that time, the better.

I would like to share an utility that I build for parallelism on PHP if that's ok.
For anyone if it is useful, [Parallel SDK](https://github.com/hschimpf/parallel-sdk) is a layer on top of parallel extension.

mnapoli:
FYI we merged the PR for ZTS in Bref layers.

To be clear, ZTS is not enabled in Bref layers, but users can rebuild runtimes with it if they wish to.

hschimpf:
Since branch `v2` was merged to `master`, I made another PR #431 on the `master` branch

Issue #14, opened 2020-01-26
Write better instructions how to contribute


Nyholm:
- How to build layers?
- How to test layers?
- How to confirm that every thing is working?

sabmeua:
I believe I almost understood the process to contribute, so I wrote my understands #17.
Also I would appreciate it if you point out when I make weird wording.

Nyholm:
Closed by #17

PR #430, opened 2023-04-07
Updated for arm support

hmimthiaz:
- Updated the project to work with x86 and arm support
- Checked existing layers and moved non working to layers-build-failed, layers-test-failed
- Updated publish and publish-docker-images, need to test it. 
- Crosschecked if x86 is working fine. Could not test fully as on my m1 mac is is awfully slow.


hmimthiaz:
Moved failed layers back 
Fixed failing and arm support for 

- gmp
- amap
- oci8
- relay

Need to work on arm support for 

- blackfire
- newrelic
- spx
- sqlsrv
- tideways

I think we can drop support for cassandra as it does not have libraries for php 8 and  arm

hmimthiaz:
Added cpu support in config

- Now make file will work automated for both arch x86 and arm.
- Checked all layers made sure all of them have 80,81,82 support.
- Sqlsrv does not have arm support
- Not very sure about Blackfire, leaving arm support for the contributor to handle it.

layers/relay/Dockerfile
```diff
@@ -1,12 +1,18 @@
...
-FROM bref/build-php-$PHP_VERSION:2 AS ext
+FROM bref/${CPU_PREFIX}build-php-$PHP_VERSION:$BREF_VERSION AS ext
 
 ARG RELAY_VERSION=0.6.0
 
+ENV LD_LIBRARY_PATH=/usr/lib:/usr/lib64:$LD_LIBRARY_PATH
+
 # Docs: https://relay.so/docs/1.x/installation#manual-installation
 
+# Added openssl11-libs
```
GrahamCampbell, line 13: Not sure this is a good idea. We don't want openssl stuff of a different version to in the main image?

GrahamCampbell (changes requested):

layers/relay/Dockerfile
hmimthiaz: The copy-dependencies script was failing due to missing library. Was not sure how to fix it. 

> [ext 5/7] RUN php /bref/lib-copy/copy-dependencies.php /tmp/relay.so /tmp/extension-libs:
#14 0.507
#14 0.507 Fatal error: Uncaught RuntimeException: This library is a dependency for /tmp/relay.so but cannot be found by 'ldd':
#14 0.507 	libssl.so.1.1 => not found

>bash-4.2# ldd /tmp/relay.so
ldd: warning: you do not have execution permission for `/tmp/relay.so'
	linux-vdso.so.1 (0x0000ffff96b83000)
	libpthread.so.0 => /usr/lib64/libpthread.so.0 (0x0000ffff96a57000)
	libssl.so.1.1 => not found
	libcrypto.so.1.1 => not found
	libzstd.so.1 => /opt/lib/libzstd.so.1 (0x0000ffff96956000)
	liblz4.so.1 => /usr/lib64/liblz4.so.1 (0x0000ffff96925000)
	libanl.so.1 => /usr/lib64/libanl.so.1 (0x0000ffff96904000)
	libc.so.6 => /usr/lib64/libc.so.6 (0x0000ffff9677e000)
	/lib/ld-linux-aarch64.so.1 (0x0000ffff96b55000)

GrahamCampbell: or we could raise an issue upstream with them?

hmimthiaz: Removing is also a good option. If someone needs it they can always do a PR. My main focus was to make the whole bunch arm compatible and I don't use any of the extension except for GD and imagick. 

cassandra and sqlsrv doesn't have arm support

GrahamCampbell: Probably we should make a separate PR to remove/fix any broken extensions we discover?

GrahamCampbell: 👌

robchett:
@mnapoli is there anything we can help with to move this forward?

gazzoy:
+1

gazzoy:
@mnapoli @GrahamCampbell any plans to supporting this?

mnapoli:
@gazzoy no plans to support ARM for AL2 as it's been made obsolete by AL2023.

So now the most sensible course of action is: AL2023 support with ARM, and then have extensions support AL2023 with ARM.

However I want to point to #526 for a more general discussion.

gazzoy:
> @gazzoy no plans to support ARM for AL2 as it's been made obsolete by AL2023.
> 
> So now the most sensible course of action is: AL2023 support with ARM, and then have extensions support AL2023 with ARM.
> 
> However I want to point to #526 for a more general discussion.

Thanks for your reply. Really makes sense. Will wait for a community-driven layer or something, then.

I really appericiate the hard work you have been made of, btw.

PR #431, opened 2023-04-19
Added `parallel` extension

hschimpf:
## Extension reference
- https://php.net/parallel

## Dependencies
- AWS Lambda PHP layer compiled with ZTS flag enabled, please refer to https://github.com/brefphp/aws-lambda-layers/pull/63

Current `CI / Build and test layer` will fail because this extension needs PHP with ZTS enabled.


mnapoli:
I think we would need to disable the build and publication of layers for this layer? (because it requires ZTS)

Issue #428, opened 2023-04-03
Mongo extension is not loaded


dcorrea777:
I have the following dockerfile

```dockerfile
FROM bref/php-82-fpm-dev

COPY --from=bref/extra-mongodb-php-82 /opt /opt

```
After I compile my dockerfile and run it from the command line, I get the following warning:

```bash
PHP Warning:  PHP Startup: Unable to load dynamic library 'mongodb.so' (tried: /opt/bref/extensions/mongodb.so (libssl.so.1.1: cannot open shared object file: No such file or directory), /opt/bref/extensions/mongodb.so.so (/opt/bref/extensions/mongodb.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
PHP 8.2.4 (cli) (built: Apr  3 2023 09:43:33) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.4, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.4, Copyright (c), by Zend Technologies
```

When I run the command `ls -la` in the path `/opt/bref/extensions` the mongodb.so file is there.

```bash
bash-4.2# ls -la  /opt/bref/extensions/
total 15788
drwxr-xr-x 1 root root    4096 Apr  1 14:17 .
drwxr-xr-x 1 root root    4096 Apr  1 14:17 ..
-rw-r--r-- 1 root root  632968 Apr  3 09:44 apcu.so
-rw-r--r-- 1 root root  465600 Apr  3 09:44 blackfire.so
-rwxr-xr-x 1 root root 1969928 Apr  3 09:44 intl.so
-rw-r--r-- 1 root root 9435544 Apr  1 14:17 mongodb.so
-rwxr-xr-x 1 root root 1160880 Apr  3 09:44 opcache.so
-rwxr-xr-x 1 root root   37600 Apr  3 09:44 pdo_mysql.so
-rwxr-xr-x 1 root root   58368 Apr  3 09:44 pdo_pgsql.so
-rw-r--r-- 1 root root 2376304 Apr  3 09:44 xdebug.so

```
But for some reason it doesn't find that extension in that path

When I run the `php -m` command it also doesn't show the mongo extension being loaded.

```
bash-4.2# php -m
[...]

PR #432, opened 2023-04-19
Mongo extension is not loaded #428

lanlin:


mnapoli:
Do you know why the tests did not fail before? Is there any way to add a test to reproduce the issue and ensure it never happens again?

I'm also curious why no one else has the issue and the tests are passing on master, without a failing test I'm not 100% sure we are fixing anything actually 🤔

PR #445, opened 2023-05-21
Fixed broken amqp extension

GrahamCampbell:
The most recent build failed, and on further investigation, the extension on pecl only works on PHP 8.0, not 8.1 and 8.2 - we need to build from source.


GrahamCampbell:
Actually fixed now. :trollface:

mnapoli:
🙌

PR #448, opened 2023-05-22
Fix the relay extension

mnapoli:


GrahamCampbell:
Did you raise with with the relay team btw?

tillkruss:
AL2 is built with CentOS7 right?

GrahamCampbell:
I think it's roughly related, yes.

tillkruss:
@GrahamCampbell: Why don't we build one specifically for AL2. What's the base image I should use?

GrahamCampbell:
AL2 is not a suitable base image. We are AL2 but not with AL2 openssl.

GrahamCampbell:
I'd probably vote for removing this extension from bref, since it is propitiatory.

tillkruss:
We can easily build an artifact, I just need to know the base image to use.

GrahamCampbell:
Like I said, the base image is custom, as per the link in the other discussion.

mnapoli:
Yeah the thing with AL2 is that it ships with outdated stuff (like OpenSSL), it's a real mess.

So to compile PHP (for Bref runtimes), we compile a lot of system libs ourselves, including OpenSSL (3).

Our base image for Bref is https://hub.docker.com/r/bref/build-php-80 (change `-80` with 81 or 82 for other PHP versions). That image includes all the libs used by Bref. But would it make sense to compile Relay specifically for this image @tillkruss?

tillkruss:
Yeah, I can just build binaries for Bref 👍

@GrahamCampbell Does Laravel use the same builds as Bref or different ones?

GrahamCampbell:
Different, but soon it will be similar enough. They are also upgrading to openssl 3.0.8, in an identical way to the way bref is building it: https://github.com/laravel/vapor-php-build/pull/96

GrahamCampbell:
Oh yeh, and note that the information int he PR description of the linked PR is largely wrong - they are not using openssl 3.1.0 - they are using 3.0.8, and will be tracking whatever version is LTS (it looks like no other 3.x will be LTS in the near future, certainly not 3.1.x).

tillkruss:
And why isn't everyone switching AL2023 which already has OpenSSL 3.x and leaving AL2 alone?

GrahamCampbell:
Because Lambda does not support AL2023 and AWS have not replied to my question: https://github.com/aws/aws-lambda-base-images/issues/92.

mnapoli:
As discussed with Till, I'll merge this PR for now to fix the build and the extension.

We can always move later to a custom-built `relay.so`.

Issue #456, opened 2023-05-31
[Datadog] PHP Startup: Unable to load dynamic library 'ddappsec.so'


annuh:
I'm running into an issue when enabling the datadog extension. When I add this extension to my Lambda's layers, the Lambda function produces the following warning:

```
Warning: PHP Startup: Unable to load dynamic library 'ddappsec.so' (tried: /opt/bref/extensions/ddappsec.so (/opt/bref/extensions/ddappsec.so: cannot open shared object file: No such file or directory), /opt/bref/extensions/ddappsec.so.so (/opt/bref/extensions/ddappsec.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
```

My full `serverless.yml` looks like this
```yaml
service: php-test

provider:
    name: aws
    #architecture: arm64
    region: eu-central-1
    environment:
        # Datadog specific settings
        DD_SITE: datadoghq.com
        DD_API_KEY: __SECRET__
        DD_TRACE_ENABLED: true
        DD_SERVICE: php-test

plugins:
    - ./vendor/bref/bref
    - ./vendor/bref/extra-php-extensions

functions:
    test:
        handler: App\Infrastructure\Bref\EventHandler
        description: 'Test function'
        runtime: php-82
        timeout: 60
        layers:
            - ${bref-extra:gd-php-82}
            - ${bref-extra:datadog-php-82}
            - arn:aws:lambda:eu-central-1:464622532012:layer:Datadog-Extension:43 # changing the layers order, didn't help

package:
    patterns:
        # Excluded files and folders for deployment
        - '!assets/**'
        - '!node_modules/**'
        - '!public/build/**'
        - '!tests/**'
        - '!var/**'
        # If you want to include files and folders that are part of excluded folders, add them at the end
        - 'var/cache/prod/**'
        - 'public/build/entrypoints.json'
        - 'public/build/manifest.json'
```

I see the function's logs in Datadog (probably because of the Datadog-Extension layer?), but there is no trace information.

Do you have any idea what's the issue here?

PR #457, opened 2023-05-31
Add ddappsec.so to datadog extension

annuh:
Fixes https://github.com/brefphp/extra-php-extensions/issues/456.

Without this I'm getting the following errors:
```
Warning: PHP Startup: Unable to load dynamic library 'ddappsec.so' (tried: /opt/bref/extensions/ddappsec.so (/opt/bref/extensions/ddappsec.so: cannot open shared object file: No such file or directory), /opt/bref/extensions/ddappsec.so.so (/opt/bref/extensions/ddappsec.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
```

### DIFF
--- a/layers/datadog/Dockerfile
+++ b/layers/datadog/Dockerfile
@@ -15,6 +15,7 @@ RUN php datadog-setup.php --php-bin=all --enable-profiling
 
 RUN cp "$(php-config --extension-dir)/ddtrace.so" /tmp/ddtrace.so && \
     cp "$(php-config --extension-dir)/datadog-profiling.so" /tmp/datadog-profiling.so && \
+    cp "$(php-config --extension-dir)/ddappsec.so" /tmp/ddappsec.so && \
     cp "$(php-config --ini-dir)/98-ddtrace.ini" /tmp/ext.ini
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell (changes requested):
I don't think we should fix this here, and instead wait for the upstream fix to be completed.

annuh:
> I don't think we should fix this here, and instead wait for the upstream fix to be completed.

Based on that ticket (https://github.com/DataDog/dd-trace-php/issues/2014), it's not clear to me what exactly will be fixed. Maybe this fix (or: work-around) will still be needed, because the datadog-php-trace setup script assumes `ddappsec.so` is always present, in a normal installation.

Nyholm:
Hm... I think this is reasonable.
Anything happened for DataDog the past few months?

annuh:
> Hm... I think this is reasonable. Anything happened for DataDog the past few months?

Not as far I know. 
The Datadog docs still mention that “the AppSec extension loads shortly at startup”.  Also, the `Unable to load dynamic library 'ddappsec.so'` warning is still logged.

PR #460, opened 2023-06-12
Fix #459 Support projects using ESM

### LINKED ISSUE #459: Module is not defined in ES module scope
Hi, I'm having the following issue: 
Using the lastest version of Laravel and Vite, with Bref, Laravel-bridge and extra-php-extensions.
The package.json file has "type": "module", and that works ok when doing "npm run build", but when trying to "serverless deploy" I got the following error:

Error:
ReferenceError: module is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/Users/x/Sites/bref-test/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///Users/x/Sites/bref-test/vendor/bref/extra-php-extensions/index.js:90:1 <- _I think it was line 74 before I touched it using prettier, the "problem" I think is module.exports = ServerlessPlugin;_
    at ModuleJob.run (node:internal/modules/esm/module_job:192:25)
   
I was trying to find out what the root of the issue is and I think this is the case
Also note that TypeScript will allow you to mix CJS and ESM, however Node.js will not. So you cannot use require() and import() in the same file, and you cannot use export default and module.exports in the same file. From this thread: https://github.com/orgs/vercel/discussions/1225#discussioncomment-4642381

If I remove "type": "module" from the package.json serverless deploy works as expected. But Vites npm run build fails when it finds the first "export default {" in postscss.config.js

I don't know what the ideal solution would be for this, anyway let me know if this is a pertinent issue, if not go ahead a close it.

Thanks!

reggie84:


mnapoli:
Thank you!

PR #485, opened 2023-09-11
Update elastic to v1.8.4

mfadul24:
Elastic agent 1.8.1 is quite buggy. I've updated it to 1.8.4.
Also I've added the required src files, in order the make the implementation work. 
See https://www.elastic.co/guide/en/apm/agent/php/current/setup.html#build-from-source

With out this src files, elastic agent error out.


Nyholm (approved):
Awesome. Thank you

PR #496, opened 2023-10-04
Upgrade to redis 6.0.1

GrahamCampbell:
It is really necessary to hard code the versions, always, otherwise there's no way to trigger new builds.


Nyholm (approved):
Thank you

PR #506, opened 2023-11-30
Updated layers.json

mnapoli:


mnapoli:
we really need to fix the release process, I'm not sure what's wrong today

PR #511, opened 2023-12-12
Add support for xmlrpc

hungnv-sr:
I know xmlrpc is deprecated and it's also not secured. However, someone (me included) might need it.
Tested on Lambda as well.

Supported version: 8.0, 8.1, 8.2, 8.3


Nyholm (approved):
Thank you

PR #514, opened 2023-12-19
update bsdiff to make it available for PHP 8.3

deminy:
This is to build the`bsdiff` layer of PHP 8.3.

The extension is compatible with PHP 8.3, as tested via [CI jobs](https://github.com/deminy/php-bsdiff/blob/master/.github/workflows/ci.yml) on GitHub.


Nyholm (approved):
Thank you

PR #515, opened 2024-01-05
Build gd for PHP 8.3

1ed:


1ed:
@mnapoli could you please merge this and the other PHP 8.3 PRs? Thanks!

ddeboer:
Yes, would be nice to have all the PHP 8.3 PRs merged to allow devs to upgrade their projects to 8.3.

@mnapoli @Nyholm Please let us know how we can help getting these PRs merged. Perhaps an extra maintainer with merge rights? Or automatically merging and tagging small PRs such as this?

mnapoli:
I'm torn, I'm not sure we actually want more things maintained here.

This repo is a collection of partially broken things, the promise for users is fairly low unfortunately. (because it's impossible to maintain all these extensions)

I'm wondering if we should just encourage people to have repos for the extensions they are actually maintaining instead of giving a false promise to users.

Anyway, that's a sidebar…

I cannot merge this today, the release process takes too much manual time and I won't have time to complete it today. I'll try tomorrow 👍

ddeboer:
> I'm wondering if we should just encourage people to have repos for the extensions they are actually maintaining instead of giving a false promise to users.

Makes sense. Should the Bref docs then be updated to describe how to build your own extension layer?

Nyholm:
I would be happy to continue the discussion, Maybe start an issue. 

I'll merge this and prepare a release.

PR #521, opened 2024-01-17
feat(imagick): make `convert` available in layer

annuh:
Running `convert` directly does not work.

For example:
```php
exec('convert --version');
```

triggers this error:
```
sh: convert: command not found
```

This PR makes `convert` available in the Imagick layer.

### DIFF
--- a/layers/imagick/Dockerfile
+++ b/layers/imagick/Dockerfile
@@ -80,6 +80,7 @@ FROM scratch
 
 # Copy things we installed to the final image
 COPY --from=ext /tmp/gs /opt/bin/gs
+COPY --from=ext /tmp/convert /opt/bin/convert
 COPY --from=ext /tmp/imagick.so /opt/bref/extensions/imagick.so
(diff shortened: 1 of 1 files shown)


Nyholm:
Could you tell me why you need `convert`?

Nyholm:
Friendly ping

annuh:
I'm sorry for the late reply! 

My initial though was, if the Imagick Bref extension includes Ghostscript (`gs`), why not also include `convert`.
In one of our serverless applications, we used to invoke `convert` directly, but I didn't realize we could replace this with the PHP Imagick extension. 🙂 For example

```diff
-exec("convert compare -metric AE '{$image1}' '{$image2}' NULL:", $output);
-$similarity = (float) $output[0];
+$similarity = (new \Imagick($image1))->compareImages(new \Imagick($image2), \Imagick::METRIC_ABSOLUTEERRORMETRIC)[1];
```

After these changes, we don't need the `imagick` binary anymore. But maybe it can be a useful for other use cases?
If you think it's not needed, I'll close this PR 🙂

PR #530, opened 2024-02-17
Use non-beta SQLSRV 5.12.0 for PHP 8.1 and later

chrishanline:
Microsoft released the non-beta version of `5.12.0` of SQLSRV on January 31, 2024. https://learn.microsoft.com/en-us/sql/connect/php/release-notes-php-sql-driver?view=sql-server-ver16

SQLSRV 5.12.0 drops support for PHP 8.0. This PR continues to use `5.11.1` for PHP 8.0, and upgrades 8.1, 8.2 and 8.3 to use `5.12.0`. Previously 8.3 was using `5.12.0beta1`


Nyholm (approved):
Great. Thank you

PR #547, opened 2024-07-12
Add brotli extension support

otezz:
Add [Brotli](https://pecl.php.net/package/brotli) support.

Since neither AWS HTTP API Gateway nor CloudFront has the ability to compress responses with Brotli, we can achieve this by using something like https://github.com/open-southeners/laravel-vapor-response-compression


GrahamCampbell:
I think this was an intentional decision since the compute requirements for compression are very high, an order of magnitude more than for gzip, and the average compression ratio is barely better.

otezz:
[They support payload compression](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-enable-compression.html) on REST API but not in HTTP API (yet). I think it should be okay to give users the option to handle the compression themselves.

Nyholm (approved):
I understand that this dont always makes sense. But I am happy to merge this.

PR #573, opened 2024-12-08
Add PHP 8.4 support for pgsql

rafaelbernard:
- **Add PHP 8.4 support for pgsql**


.gitignore
```diff
@@ -1,3 +1,6 @@
 /vendor/
 composer.lock
 .php_cs.cache
+
+# IDEs
+.idea
```
Nyholm, line 6: This is a hard no for two reasons. 
- It belongs in your global gitignore
- Jetbrains actually recommends committing this. 

So no matter the philosophy you have about this directory, you should not add it into a project.

Nyholm (changes requested):
Thank you for this PR. 

Just remove your changes in .gitignore and I'll be happy to merge.

.gitignore
rafaelbernard: Thanks for providing the details. That is all good.

Nyholm (approved):
Great. Thank you

rafaelbernard:
Thanks for reviewing and merging.

PR #578, opened 2024-12-25
Build for PHP 8.4 and update libraries

mu1f407:
Hi, this PR adds a build for PHP 8.4 for all extensions that are currently built for 8.3.

The versions of the libraries are also updated to the latest ones. In some cases the update is over the major version.

We will see how the CI tests turn out. If there are any problems, I will take the specific extension out of PR and it will be handled separately later (by me or someone else).


mnapoli:
Thanks! It seems like a few extensions failed, might as well remove them from the list for now?

mu1f407:
Update summary:

**PHP 8.4 not supported by upstream project yet:**
elastic-apm
newrelic

**Build fails, should be handled in separate PRs:**
imagick
imap (moved to PECL in 8.4)
memcached
rdkafka
swoole

**Successfully updated for PHP 8.4, but constrained by Amazon Linux 2 in some way:**
amqp (old cmake for rabbitmq-c >=0.14; stick to 0.13)
oci8 (old glibc for Oracle Instant Client >=23; stick to 21.x)

**Successfully updated for PHP 8.4:**
blackfire
bsdiff
calendar
datadog (library major version update 0.96.0 -> 1.5.1)
decimal
ds
excimer
gmp
gnupg
grpc (not sure if it should take more than 10 minutes to compile for each PHP version)
h3
igbinary
ldap
mailparse
maxminddb
msgpack
odbc-snowflake (library major version update 2.25.0 -> 3.5.0)
opentelemetry
pcov
redis
redis-igbinary
relay
scoutapm
snmp
ssh2
symfony-runtime
tideways
tidy
uuid
xdebug
xlswriter
xmlrpc
yaml

layers/amqp/Dockerfile
```diff
@@ -19,7 +21,7 @@ RUN cmake --build . --target install
 WORKDIR ${AMQP_BUILD_DIR}
 RUN git clone https://github.com/php-amqp/php-amqp
 WORKDIR ${AMQP_BUILD_DIR}/php-amqp
-RUN git reset --hard 618e06ad2ef867598831cdd3faadba0dd65be917
+RUN git checkout v2.1.2
```
mnapoli, line 24: Could you explain this change?

layers/datadog/Dockerfile
```diff
@@ -7,15 +7,15 @@ ENV DDTRACE_BUILD_DIR=${BUILD_DIR}/ddtrace
...
+      https://github.com/DataDog/dd-trace-php/releases/download/1.5.1/datadog-setup.php
 
 WORKDIR ${DDTRACE_BUILD_DIR}
 
 RUN php datadog-setup.php --php-bin=all --enable-profiling
 
 RUN cp "$(php-config --extension-dir)/ddtrace.so" /tmp/ddtrace.so && \
     cp "$(php-config --extension-dir)/datadog-profiling.so" /tmp/datadog-profiling.so && \
-    cp "$(php-config --ini-dir)/98-ddtrace.ini" /tmp/ext.ini
+    cp "$(php-config --ini-dir | cut -d: -f1)/98-ddtrace.ini" /tmp/ext.ini
```
mnapoli, line 18: Could you explain this change?

layers/oci8/Dockerfile
```diff
@@ -10,27 +10,28 @@ ENV ORACLE_BUILD_DIR=${BUILD_DIR}/oracle
...
 
+# Instant Client newer than 21.x requires glibc 2.28+ which is not available on Amazon Linux 2
 RUN mkdir -p ${ORACLE_BUILD_DIR}; \
     cd ${ORACLE_BUILD_DIR}; \
-    curl -o oci-basic.zip https://download.oracle.com/otn_software/linux/instantclient/2110000/instantclient-basiclite-linux.x64-21.10.0.0.0dbru.zip && \
-    unzip oci-basic.zip -d src && \
-    curl -o oci-sdk.zip https://download.oracle.com/otn_software/linux/instantclient/2110000/instantclient-sdk-linux.x64-21.10.0.0.0dbru.zip && \
-    unzip oci-sdk.zip -d src
+    curl -o oci-basic.zip https://download.oracle.com/otn_software/linux/instantclient/2116000/instantclient-basiclite-linux.x64-21.16.0.0.0dbru.zip && \
+    unzip oci-basic.zip -d src -x META-INF/* && \
```
mnapoli, line 17: Could you explain this change?

mnapoli (commented):
Thanks a lot! I added some questions inline to make sure I understand everything.

Some notes regarding the Datadog migration from 0.x to 1.x: https://github.com/DataDog/dd-trace-php/blob/1.0.0/UPGRADE-1.0.md (for those stumbling on this PR in the future).

layers/amqp/Dockerfile
mu1f407: This commit referenced php-amqp/php-amqp@618e06a, a fix for PHP 8.1. I assume this was needed back then when this commit was not tagged yet. Tag v2.1.2 is from Jan 22, 2024 and this fix is already contained in it. So this is basically an update of php-amqp source to latest release.

layers/datadog/Dockerfile
mu1f407: The `php-config --ini-dir` returns `/opt/bref/etc/php/conf.d:/var/task/php/conf.d`. This `cp` command assumes that there is only single path in the output, otherwise the source file is not found. Probably, there was some change in build-php image recently, which changed the behavior and now contains two paths separated by colon. If you have better idea for fix, I would be happy to test and implement it.

layers/oci8/Dockerfile
mu1f407: The `-x META-INF/*` excludes this directory from unzip extraction. Seems like it was not there previously and added at some time. These are some metadata like checksums of files, not needed for compilation. The issue is that both oci-basic.zip and oci-sdk.zip are extracted to single output directory `src`. Because there are equally named files in this directory, the unzip of the latter file causes interactive propmt: `replace src/META-INF/MANIFEST.MF? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
(EOF or read error, treating as "[N]one" ...)` and the unzip fails with error.

mnapoli:
Awesome, thanks for the quick responses!

.github/workflows/blackfire.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.4
+          php-version: 8.4
```
TimWolla, line 16: This likely broke the blackfire.yml workflow: https://github.com/brefphp/extra-php-extensions/actions/runs/13669368623/job/38216533337

The “Publish” step fails with:

```
PHP Fatal error:  Uncaught TypeError: curl_multi_setopt(): Argument #2 ($option) must be a valid callback for option CURLMOPT_PUSHFUNCTION, no array or string given in /home/runner/work/extra-php-extensions/extra-php-extensions/vendor/symfony/http-client/Internal/CurlClientState.php:57
Stack trace:
#0 /home/runner/work/extra-php-extensions/extra-php-extensions/vendor/symfony/http-client/Internal/CurlClientState.php(57): curl_multi_setopt()
#1 /home/runner/work/extra-php-extensions/extra-php-extensions/vendor/symfony/http-client/Internal/CurlClientState.php(83): Symfony\Component\HttpClient\Internal\CurlClientState->reset()
#2 [internal function]: Symfony\Component\HttpClient\Internal\CurlClientState->__destruct()
#3 {main}
  thrown in /home/runner/work/extra-php-extensions/extra-php-extensions/vendor/symfony/http-client/Internal/CurlClientState.php on line 57
```

which is likely due to `symfony/http-client` being pinned to 5.3.8 in composer.json.

PR #586, opened 2025-03-05
Use official `ghcr.io/tideways/php` docker image for Tideways installation

TimWolla:
Note: I assume this works, but I don't know about bref and thus couldn't end-to-end test it. I verified locally that the Docker image still builds and that the result is identical using [diffoci](https://github.com/reproducible-containers/diffoci).

--------------

The provided tooling will do the heavy lifting of choosing the correct `.so` file for a given PHP version, it will automatically support other architectures and downloading and verification is handled by Docker instead of a manual `curl` command.


mnapoli:
Thanks!

PR #592, opened 2025-04-28
Add new layer for MongoDB driver for PHP v2.0.0

### LINKED ISSUE #591: Support for `mongodb` 2.0.0 php driver
Hi, 

The mongodb team released a 2.0.0 version of the [MongoDB driver for PHP](https://github.com/mongodb/mongo-php-driver/releases) (including a version 2.0.0 for the [MongoDB PHP library](https://github.com/mongodb/mongo-php-library/releases/tag/2.0.0)), which I want to use in my project. 

I can make a PR for a new version (or just a re-run of the build, because the pecl command does not explicit requires a version), but I think that will potentially break backward-compatibility with other users that need the 1.x version of the php driver. Im only using this driver in one project, so for me the BC is not a big deal, but it might be for others.

My idea is to create a new bref extension `mongodb2`, to support both 1.x and 2.x versions.

Any advice from @mnapoli or @Nyholm ?

starred-gijs:
Fixes #591

- Created new layer for MongoDB driver for PHP **v2.0.0**, with minimum PHP version is 8.1, see [Language Compatibility](https://www.mongodb.com/docs/drivers/php-drivers/#language-compatibility)
- Needed to pin the existing layer to v1.21.0 (latest v1) beause it would install v2


Nyholm (approved):
Thank you

PR #612, opened 2025-07-28
Update datadog extension to support php 84

Gaitholabi:
The layer:php-84-fpm currently causes a crash in the Datadog extension due to profiling being enabled by default.
- https://github.com/DataDog/dd-trace-php/releases/tag/1.5.1

 Since the Datadog extension in the layer is pretty outdated, this PR updates it to the latest available version to resolve the crash and benefit from the latest improvements.


Gaitholabi:
Hey @mnapoli - Thank you for merging this promptly. Is there anything I could help with to get this change released for a useable layer?

Gaitholabi:
The intention was only to see if I could give a hand with it, not to take out of your time. Thanks for the reply.

PR #658, opened 2026-04-09
fix(imagick): add libtiff-devel for TIFF decode support

imOmerg:
In v1 (AL2), the dnf install line included ImageMagick-devel which transitively pulled in libtiff-devel. In v3 (AL2023), ImageMagick-devel was dropped and libtiff-devel was never added explicitly.


mnapoli:
thanks!

PR #664, opened 2026-06-02
Update Datadog layer to use dd-trace-php v1.20.0

ZachAtLevno:
Bump Datadog layer to update `dd-trace-php` to `v1.20.0` for latest bugfixes.

Layer builds and test passes on all PHP versions.
```bash
layer=datadog make test
#> ...
#> - Test passed
```

### DIFF
--- a/layers/datadog/Dockerfile
+++ b/layers/datadog/Dockerfile
@@ -7,7 +7,7 @@ ENV DDTRACE_BUILD_DIR=${BUILD_DIR}/ddtrace
 RUN set -xe; \
     mkdir -p ${DDTRACE_BUILD_DIR}; \
     curl -Ls -o ${DDTRACE_BUILD_DIR}/datadog-setup.php \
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks!

mnapoli:
Unfortunately I will have to rollback because the layer now exceeds the max layer size: https://github.com/brefphp/extra-php-extensions/actions/runs/28175372059/job/83466160258

ZachAtLevno:
Hi @mnapoli 

Thanks for taking the time to try this. It looks like the layer is too big to be uploaded as base64, not too big to be a layer. I suspect the layer upload method will need to be changed to be S3-based, rather than base64-encoded content. See below:
https://github.com/brefphp/extra-php-extensions/blob/e2ef5a7ddc0ddd2e0f8455d46e857e8f993ce8cb/src/Aws/LayerPublisher.php#L48
https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-lambda-2015-03-31.html#publishlayerversion

Let me know if I can help further.

PR #665, opened 2026-06-18
relay: latest version by default, arm64-ready, build on AL2023 + OpenSSL 3

tillkruss:
Modernizes and enables the **Relay** layer.

### Changes
- **Latest version by default** — `RELAY_VERSION` defaults to empty and resolves the latest stable from `builds.r2.relay.so/meta/latest` (was pinned to `0.9.1`); still overridable via `--build-arg RELAY_VERSION=x.y.z`.
- **arm64-ready** — Relay's arch suffix is derived from `uname -m` (`x86-64` / `aarch64`) instead of hard-coding `x86-64`.
- **Amazon Linux 2023 / OpenSSL 3** — base image bumped to `:3`, switched to Relay's self-contained `el9` **`relay-pkg.so`** build (links system OpenSSL 3 + glibc 2.34), **removing the custom OpenSSL 1.1 compile** — the slowest build step.
- **Enabled the layer** — `config.json` was empty (so it was never built/published); now `["82", "83", "84", "85"]`.

### Tested in Docker (PHP 8.3 & 8.5)
- `make test` — extension loads, `\Relay\Relay` available, clean `php -v`.
- Live Redis connection: plain TCP **and** a verified TLS handshake (exercises OpenSSL 3) — `PING`, `SET`/`GET`, `INCR` all round-trip.


szepeviktor (approved):

layers/relay/Dockerfile
```diff
@@ -1,13 +1,14 @@
 # syntax = docker/dockerfile:1.4
 ARG PHP_VERSION
 ARG BREF_VERSION
-FROM bref/build-php-$PHP_VERSION:2 AS ext
+FROM bref/build-php-$PHP_VERSION:3 AS ext
 
-ARG RELAY_VERSION=0.9.1
+# Leave empty to install the latest stable release
+ARG RELAY_VERSION=
```
mnapoli: Can you specify a version number here? Else the layer is never rebuilt and the extension is never updated.

tillkruss: Done!

mnapoli:
Thanks!

PR #669, opened 2026-07-24
Add ssh2 layer for PHP 8.5 (ssh2-php-85)

zmalter99:
ssh2 is only published up to php-84, so deploying an 8.5 function with `${bref-extra:ssh2-php-85}` fails with `Unknown Bref extra layer named "ssh2-php-85"`.

This adds `"85"` to `layers/ssh2/config.json`, same as gd and the other extensions that already build for 8.5.


mnapoli:
Thanks!

PR #675, opened 2026-08-17
Add PHP 8.5 support to pgsql

alejandrohernandez-aircury:
This PR adds PHP 8.5 support for the pgsql extension.


alejandrohernandez-aircury:
Closing PR because of this https://bref.sh/docs/upgrading/v3#postgresql-extension-is-enabled-by-default I will investigate why is not working in my application with v3

mnapoli:
@alejandrohernandez-aircury yes let me know. Please note that `pdo_pgsql` is built-in, I suspect pgsql is different?

